PR #105, opened 2013-11-14
Debugging and optimization

esc:
some more upgrades to the debugging and optimization chapter. Probably some are controversial, so will probably have to discuss some things here. Especially the section on Big-Oh notation is perhaps a bit much. However the ideas of a single for loop being cheaper than two are probably worth to communicate.


vahtras:
I think it is time to close this PR from 2013, which by now also has merge conflicts. Better start fresh.

PR #120, opened 2014-08-22
[MRG-1] Warning about issues with Python3

quantummechanist:
Include workaround information for users that have made the prudent switch to Python3


GaelVaroquaux:
Thanks a lot for the PR. 

However, I don't understand: I have tested, and under Python 3, scipy.misc.imread works perfectly.

@quantummechanist  Maybe this wasn't the case when you wrote this PR? But I think that, in the current state of affairs, we shouldn't merge this PR.

GaelVaroquaux:
I am closing this, because I think that Python 3 issues have been addressed. If it is not the case, feel free to reopen.

PR #124, opened 2014-09-22
Update with minor improvements

strpeter:
- Clarified the list,
- Corrected number of features,

At some point one should mention how to write comments and what is allowed as a variable name.


GaelVaroquaux:
Merging, although I do agree that the comas feel weird (but I don't want to nitpick)

PR #128, opened 2015-01-17
Typo at line #513

rishy:
At line #513, Result of "a[2, 1]" should be "0" instead of "10".


esc:
I believe you have not understood the code example correctly. This statement *assigns* the value ``10`` to the array at position ``[2,1]``. It doesn't query it.

PR #133, opened 2015-02-25
Pylab deprecated now

matthew-brett:


esc:
I'd be in favour of pulling this like it is. BTW: is pylab mode now default? What is it deprecated in favour of now?

matthew-brett:
I believe the advice is to always use the explicit imports ('matplotlib.pyplot as plt') and the %matplotlib inline magic to show the plots inline (notebook) or %matplotlib (no 'inline') at the console.

takluyver:
The `--pylab` option still works in plain terminal IPython, but not in the notebook; notebooks created using pylab mode have no indication that you need to run them in pylab mode, which is problematic. So it's definitely preferable to use magics within the notebook. And then it's probably best to recommend those magics universally, rather than confusing people by using one mechanism in the terminal and another in the notebook.

A separate issue is that many people now frown on the way pylab stuffs a whole load of names into your namespace. We're not getting rid of `%pylab`, because it is convenient at times to have all those things quickly accessible, but we recommend teaching with `%matplotlib`, which just sets up matplotlib integration (`%matplotlib inline` for inline plots), and then doing the imports manually.

esc:
Okay so two recommendations:

1) use explicit imports
2) use the %matplotlib magic in the notebook

?

takluyver:
Yep.

Matthew reminded me that @Carreau wrote a blog post about this: http://carreau.github.io/posts/10-No-PyLab-Thanks.html

takluyver:
Oh, and the `%matplotlib` magic is also useful in the terminal - it does some clever integration stuff so that having a plot window open doesn't block your terminal, so you can carry on typing in code.

matthew-brett:
I think I covered these recommendations in the proposed changes, but happy to be corrected.

esc:
Okidokey!

GaelVaroquaux:
Thank you Matthew! This is very helpful.

fboris:
Thanks to clear explanation. Why I still see `pylab` on [matplotlib lectures](http://scipy-lectures.github.io/intro/matplotlib/matplotlib.html)?

esc:
fboris: because we haven't made a new release yet.

PR #136, opened 2015-08-27
Updated the scikit-image tutorial to be consistent with recent versions

emmanuelle:
of skimage.


packages/scikit-image/index.rst
```diff
@@ -71,30 +79,40 @@ Image = ``np.ndarray``
...
-    >>> from skimage import data, filter  # most functions are in subpackages 
+    >>> from skimage import data  # most functions are in subpackages 
     >>> from skimage.morphology import disk
 
+::
+
+    >>> try:
+    ...     from skimage import filters
+    ... except ImportError:
+    ...     from skimage import filter as filters
```
GaelVaroquaux: You should explain why you do this. It will confuse readers

packages/scikit-image/index.rst
```diff
@@ -224,26 +216,26 @@ unsigned.
...
 
 
-Utility functions are provided in :mod:`skimage.util` to convert both the
+Utility functions are provided in :mod:`skimage` to convert both the
 dtype and the data range, following skimage's conventions:
 ``util.img_as_float``, ``util.img_as_ubyte``, etc.
 
 See the `user guide
-<http://scikit-image.org/docs/0.8.0/user_guide/data_types.html>`_ for
+<http://scikit-image.org/docs/dev/user_guide/data_types.html>`_ for
```
GaelVaroquaux: I think that linking to a dev documentation is a bad idea, you could link to http://scikit-image.org/docs/dev/user_guide/data_types.html

GaelVaroquaux:
Aside from the minor comments, this looks good!

GaelVaroquaux:
Thx!

PR #154, opened 2015-08-28
make all doctest in scipy.rst work

gertingold:
In order to make doctests work, the seed of the random number generator has been set and a file fname.png is created.


GaelVaroquaux:
Merging. Thanks

Issue #137, opened 2015-08-27
Improve the CSS


GaelVaroquaux:
* Check that it renders well on mobile phone

* Use PureCSS for the above, and because it looks cooler

* Add a "table of contents" that (models are http://nilearn.github.io/introduction.html#python-for-neuroimaging-a-quick-start http://bokeh.pydata.org/en/latest/, the design of bokeh being nicer):

   - Doesn't scroll down
   - Highlights the current section
   - Disappears on small screens (mobile friendly + tutorial friendly)

NelleV:
I can work on this.

raquel-ucl:
I can help as well

GaelVaroquaux:
The sidebar is now implemented in master. Mobile work OK. Closing.

PR #155, opened 2015-08-28
#137 Now uses better fonts and font-style

NelleV:
Small css changes to have a more modern "pure.css" look.

This is part of #137


NelleV:
@GaelVaroquaux It's a go.

GaelVaroquaux:
Thx!! Merging

PR #160, opened 2015-08-29
make (almost all) doctests pass in mathematical_optimization

gertingold:
Some remarks:
* In contrast to the previous behavior, the example for simulated annealing works just fine for the seed 
   chosen. If the example should not behave so well, may be another seed needs to be chosen.
* In line 954, doctest complained about `unexpected EOF while parsing` even though everything seemed 
   syntactically correct. As a temporary solution, I removed the linebreak, but this should in principle not be
   necessary.
* A problem remains in line 995 where some of the output of subroutine COBYLA is not considered as output
   by doctest. It could be removed to make the doctest work, but on the other hand, the user will have this 
   output displayed when running the example. Therefore, it probably should be presented in the lecture notes. 
   Replacing it by an ellipsis also does not seem very helpful.

### DIFF
--- a/advanced/mathematical_optimization/index.rst
+++ b/advanced/mathematical_optimization/index.rst
@@ -951,8 +951,7 @@ as box bounds can be rewritten as such via change of variables.
 
     >>> def f(x):
     ...    return np.sqrt((x[0] - 3)**2 + (x[1] - 2)**2)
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
This is great! Merging

PR #161, opened 2015-08-29
Numpy doctests

GaelVaroquaux:
Fix numpy doctests


GaelVaroquaux:
Not complete fix, but still progress

PR #168, opened 2015-08-29
remove use of anneal and add note on its deprecation

pdebuyl:
note added in the scipy intro file


GaelVaroquaux:
Merging! Thanks!

PR #170, opened 2015-08-29
normalize output when a namedtuple is involved

pdebuyl:


GaelVaroquaux:
Actually, let me revert my previous view on this: this way of doing things is actually quite bad for teaching.

I suggest that we simply skip this test, as there is only one.

pdebuyl:
Ok, I force-push my master with a `+SKIP`.

GaelVaroquaux:
Merging! Thx!

PR #174, opened 2015-08-29
make two failing doctests in mathematical_optimization pass

gertingold:
After rerunning the tests, it seems that there are still two problems which this PR attempts to solve.
* In one case a random seed has to be set.
* In another case, it seems that the output of `optimize.fmin_l_bfgs_b` got lost. It is restored with this PR.


GaelVaroquaux:
Fantastic!! Merging.

We now only have only one failing file. I will try to have a look at it, as it is a bit challenging.

PR #175, opened 2015-08-30
Try to fix image_processing

GaelVaroquaux:
Working on fixing the image processing


GaelVaroquaux:
Merging, as it is an improvement.

Issue #141, opened 2015-08-27
Remove IPython style prompts


GaelVaroquaux:
Replace all the IPython style prompts by standard ">>>"

Rational:

* Simplifies build (we can remove ipython_console_highlighting.py)
* Enables doctesting
* The javascript ">>>" thing to remove interactively prompts is very cool.

PR #176, opened 2015-08-30
converted IPython prompts to regular prompts where appropriate

gertingold:
In view of #141, this PR replaces IPython prompts by regular prompts in some of the files and makes a part of the doctests work. The replacement has not been done when the functionality discussed relies on IPython. Not all doctest run ok yet.

### DIFF
--- a/advanced/image_processing/index.rst
+++ b/advanced/image_processing/index.rst
@@ -91,38 +91,34 @@ Writing an array to a file:
     :align: center
     :scale: 50
 
-.. Comment to provide lena.png so that doctest passes
-   >>> from scipy import misc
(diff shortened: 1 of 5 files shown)


GaelVaroquaux:
That's really useful!

Do you think that you could:

1) rebase / merge master, to get the fixes that are in master

2) fix the error that this has introduced (to keep with the goal of having travis always green).

gertingold:
I have taken out the commits which break doctests. These will (hopefully) be the subject of a future PR. The remaining commits pass Travis.

GaelVaroquaux:
Thanks! Merging

PR #179, opened 2015-08-31
Fix users and devel mailing lists addresses

rougier:
Update of the the matplotlib mailing lists addresses


GaelVaroquaux:
Thanks. Merging

PR #181, opened 2015-09-03
Issue 66: update authors list

vahtras:
Created a makefile rule to list all contributors to be listed at the end of AUTHORS.rst (alphabetical order on last name)

    $ make contributors
    - arunpersaud

    - Matthew Brett

    - Lars Buitinck
    ...

Possible issues:

* Some entries are only github usernames because the account holder has not provided his or her full name
* Should names be include here which are already in the "Chapter authors" section?
* All contributors are listed, even if they only made a single commit with a spell fix. That can be discussed but I have chosen not to exclude anyone here.

/Olav

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -67,8 +67,6 @@ Listed by alphabetical order
 
 - egens
 
-- emmanuelle
-
 - Andre Espaze
 
 - André Espaze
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
@vahtras : I think that this is great.

However, if we do this, we must also make sure to curate the ".mailmap" file, to fix when authors used different email addresses. I have made a pass on the mailmap file. Could you please rebase on master and regenerate the authors list in your PR.

vahtras:
OK, rebased and updated AUTHORS.rst

GaelVaroquaux:
Thank you. Merging.

PR #182, opened 2015-09-03
remove double negative in intro paragraph

jooolia:
Double negative makes the first sentence difficult to follow.


GaelVaroquaux:
That's an improvement. Thanks. Merging

PR #183, opened 2015-09-05
Added packages directory to the set of doctested files.

emmanuelle:

PR #184, opened 2015-09-05
Doctest packages

GaelVaroquaux:
Superseeds #183 and tries to fix the tests on top of it.

### DIFF
--- a/packages/sympy.rst
+++ b/packages/sympy.rst
@@ -2,6 +2,8 @@
 .. TODO: bench and fit in 1:30
 
 .. TODO: plotting <- broken in OSX
+   >>> import sympy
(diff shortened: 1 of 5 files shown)


GaelVaroquaux:
Tests pass. Merging.

packages/sympy.rst
```diff
@@ -143,7 +154,7 @@ Simplify
 Use simplify if you would like to transform an expression into a
 simpler form::
 
-    In [19]: simplify((x+x*y)/x)
+    In [19]: simplify((x + x*y)/x)
```
emmanuelle: I think pep8 would prefer ``(x + x*y) / x``

GaelVaroquaux: > I think pep8 would prefer (x + x*y) / x

Can you fix this in master, please

PR #190, opened 2015-09-13
corrected rst to restore code example

gertingold:
Following the conversion from an IPython-style to a Python-style code example, the code example ended up being interpreted as an incorrect figure caption. The code example thus effectively had disappeared. This PR corrects the problem and restores the code example.


GaelVaroquaux:
Thanks! And sorry for being slow to merge.

PR #191, opened 2015-09-13
make code example visible

gertingold:
The `literalcode` directive does not accept the keyword `start_line`. Therefore, the code example did not appear in the output. Probably the most stable way to include the code segment is the `pyobject` keyword used in this PR.


GaelVaroquaux:
Thanks! Merging

PR #194, opened 2015-09-21
[WIP] All examples

GaelVaroquaux:
Make all the examples run with the gallery runner.

In master, README.txt and titles were missing, so examples wouldn't all run.

This is a WIP as I am pretty sure that examples won't run on the continuous integration system.


GaelVaroquaux:
The tests are passing. It is simply because Travis does not build the docs, and thus does not build the examples.

We will have to enable Circle-CI to do this (as in nilearn). But this is for later. I am merging this in the mean time.

PR #196, opened 2015-09-23
Fix tutorial and scripts pylab/pyplot

rougier:
Removed pylab references (in favor of pyplot) and changed all the scripts.

I use `plt` instead of `pl`since `plt` is more standard I think.

### DIFF
--- a/intro/matplotlib/examples/plot_grid.py
+++ b/intro/matplotlib/examples/plot_grid.py
@@ -5,7 +5,6 @@ Grid elaborate
 An example displaying a grid on the axes and tweaking the layout.
 """
 
-import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib.ticker import MultipleLocator
 
--- a/intro/matplotlib/examples/plot_grid.py
+++ b/intro/matplotlib/examples/plot_grid.py
@@ -5,11 +5,12 @@ Grid elaborate
 An example displaying a grid on the axes and tweaking the layout.
 """
 
-import pylab as pl
+import numpy as np
+import matplotlib.pyplot as plt
 from matplotlib.ticker import MultipleLocator
 
-fig = pl.figure(figsize=(8, 6), dpi=72, facecolor="white")
-axes = pl.subplot(111)
+fig = plt.figure(figsize=(8, 6), dpi=72, facecolor="white")
+axes = plt.subplot(111)
 axes.set_xlim(0, 4)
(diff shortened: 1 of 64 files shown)


intro/matplotlib/examples/plot_grid.py
```diff
@@ -5,11 +5,12 @@
 An example displaying a grid on the axes and tweaking the layout.
 """
 
-import pylab as pl
+import numpy as np
```
GaelVaroquaux: Why do you need the numpy here? (I guess I am being dumb)

intro/matplotlib/examples/plot_mew.py
```diff
@@ -4,20 +4,22 @@
 
 Demo the marker edge widths of matplotlib's markers.
 """
-import pylab as pl
+
+import numpy as np
```
GaelVaroquaux: Same question: do you need numpy here?

intro/matplotlib/examples/plot_grid.py
rougier: I tried to have automatics replacement (using emacs) and this is the result...
I think there are many scripts with the same problem :(
I'll try to correct that

intro/matplotlib/examples/plot_mew.py
rougier: Nope.

intro/matplotlib/examples/plot_multiplot_ex.py
```diff
@@ -5,24 +5,25 @@
 Show multiple subplots in matplotlib.
 """
 
-import pylab as pl
+import numpy as np
```
GaelVaroquaux: And here too?

I am asking this because in the logic of being didactic, I think that not importing something not used is better (less cognitive overall).

intro/matplotlib/examples/plot_grid.py
GaelVaroquaux: > I think there are many scripts with the same problem :(

Runing pyflakes on ".py" should list the scripts.

Thanks a lot!

GaelVaroquaux:
Thanks a lot for doing this. It's pretty awesome. Keeping the notes up to date provides value for everybody.

Don't forget to add a line about this in the CHANGES.txt (which I'll need to update anyhow before the release).

GaelVaroquaux:
Thanks! Merging

PR #203, opened 2015-09-29
Fix for the PDF cover

rougier:
Small fix for the cover that uses `pdfpages` instead of `includegraphics`.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -319,6 +319,7 @@ latex_preamble = r"""
 latex_elements = {
     'fontpkg': '\\usepackage{lmodern}',
     'fncychap': r'''%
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
When I try to build your PR, I get the following LaTeX error:
<pre>
! LaTeX Error: Option clash for package graphicx.
</pre>
Are you not getting it? And if so, do you have any idea why I am getting it?

rougier:
Yes, same here. I'll try to find a solution

rougier:
Should work now.

GaelVaroquaux:
Works fine. Merging.

PR #208, opened 2015-10-01
Fix authors list

rougier:


GaelVaroquaux:
Thanks. Merging.

For the context: we need this to be able to reference the scipy lectures on academic archiving systems that do a check on the author list.

PR #214, opened 2015-10-25
Calling reload in Python3

ogeorg:
While trying the example, I found reload as explained here did not work in Python3. Added how to do it in Python3.


GaelVaroquaux:
That's great. Very useful correction. Thank you. Merging.

PR #216, opened 2015-11-03
Replace scipy.misc.lena() with scipy.misc.face()

thouis:
The lena image is no longer part of scipy.misc, as of 0.17.
Replace lena with face, using face(gray=True).

In addition, changed one reference (in
packages/scikit-image/index.rst) from data.lena to data.astronaut, for
consistency.


thouis:
In addition to scipy no longer supplying the Lena image in scipy.misc(), I think it's worth considering this patch for community-building reasons.  Many people are uncomfortable with the widespread use of the Lena image (as mentioned in the scipy docs), and it's worth the minimal effort to avoid giving unnecessary offense.

GaelVaroquaux:
But how about backward compatibility? This means that the lecture notes
will work only on 0.17, which, I believe, is not even release. This in
turn will imply that users who try to copy-paste the code, or do
exercises will have failures, and will not reproduce the examples.

Is scipy not going through a deprecation cycle? This is a backward
incompatible change.

thouis:
0.17 no longer has the lena image (referencing it raises a RuntimeError, according to the docs).

The changes here should work back to 0.12.0

GaelVaroquaux:
> The changes here should work back to 0.12.0

OK. I wasn't aware that the 'face' image was in scipy for so long. I will
review this PR tonight. Thanks a lot.

GaelVaroquaux:
@thouis : I am sorry, but the tests broke (we doctest the lectures):
https://travis-ci.org/scipy-lectures/scipy-lecture-notes/jobs/89056874

Some of the problems are because the numbers of array changed. Others are because scikit-image added astronaute too recently (we have a policy of working with old stuff).

thouis:
OK, I'll try to address the test failures tomorrow.

Suggestions for a better image to use?  If scikit-image relies on scipy, can I just use the misc.face?

GaelVaroquaux:
I say yes, and I asked @emmanuelle and she is :+1: too.

advanced/image_processing/examples/plot_display_face.py
```diff
@@ -1,26 +1,26 @@
 """
-Display lena
-=============
+Display a Face
+==============
 
-An example that displays lena with matplotlib.
+An example that displays a face with matplotlib.
```
GaelVaroquaux: Isn't this some form of badger, rather than a face?

advanced/image_processing/examples/plot_display_face.py
```diff
@@ -1,26 +1,26 @@
 """
-Display lena
-=============
+Display a Face
```
GaelVaroquaux: Isn't this some form of badger, rather than a face?

advanced/image_processing/examples/plot_display_face.py
thouis: Racoon.  Scipy calls it a racoon face.

advanced/image_processing/examples/plot_face.py
```diff
@@ -0,0 +1,14 @@
+"""
+Displaying a Face
```
GaelVaroquaux: Same remark on the badger (or animal)

advanced/image_processing/examples/plot_display_face.py
GaelVaroquaux: > > +An example that displays a face with matplotlib.

> Racoon. Scipy calls it a racoon face.

Sounds good to me. I would write a "racoon face", rather than "face".

advanced/image_processing/examples/plot_face_tv_denoise.py
```diff
@@ -2,18 +2,19 @@
 Total Variation denoising
 ===========================
 
-This example demoes Total-Variation (TV) denoising on Lena.
+This example demoes Total-Variation (TV) denoising on a Face.
```
GaelVaroquaux: raccoon face

advanced/image_processing/examples/plot_face_tv_denoise.py
```diff
@@ -2,18 +2,19 @@
...
 """
 
 import numpy as np
 import scipy
+import scipy.misc
 import matplotlib.pyplot as plt
-from skimage.filter import denoise_tv_chambolle
+from skimage.restoration import denoise_tv_chambolle
```
GaelVaroquaux: We decided to postponed that change (filter -> restoration) to still support old versions of scikit-image. In a later release of the scipy lectures, we'll change it.

advanced/image_processing/examples/plot_geom_face.py
```diff
@@ -0,0 +1,45 @@
+"""
+Geometrical transformations
+==============================
+
+This examples demos some simple geometrical transformations on a face.
```
GaelVaroquaux: raccoon face :)

advanced/image_processing/examples/plot_interpolation_face.py
```diff
@@ -2,22 +2,22 @@
 Image interpolation
 =====================
 
-The example demonstrates image interpolation on Lena.
+The example demonstrates image interpolation on a face.
```
GaelVaroquaux: raccoon face :)

GaelVaroquaux:
Aside from the minor remarks above, I went through the examples, and they are still very didactic, so this is a very very welcomed contribution. Thank you!

thouis:
Travis seems to have fallen over.  It's failing before it gets to the changes in this PR.

GaelVaroquaux:
Yes. It seems that Continuum has changed something in anaconda. Let me fix this in master, and then you can rebase on master.

GaelVaroquaux:
Done in PR #217, that is now merged to master. Rebasing on master should fix your problem.

GaelVaroquaux:
Closing as superseeded by #218

PR #220, opened 2015-12-22
Small changes on intro/scipy (mainly scipy.integrate)

Naereen:
Small changes:
- [X] Add LaTeX math when needed (...`` :math:`x` ``... role instead of double backquotes ... ``x`` ...),
- [X] Remove extra ``----`` under titles,
- [X] Change  link for the homepage of sfePy (old https://code.google.com/p/sfepy/ → now http://sfepy.org/doc/),
- [X] Finish adding maths (for [scipy.integrate](http://docs.scipy.org/doc/scipy/reference/tutorial/integrate.html#integration-scipy-integrate), ``Om`` → :math:`\Omega` in LaTeX),

And:
- [X] A complete review and check (I think there is no other typo)...

### DIFF
--- a/intro/scipy.rst
+++ b/intro/scipy.rst
@@ -777,30 +777,32 @@ so the system will be underdamped, because::
 For the :func:`scipy.integrate.odeint` solver the 2nd order equation 
 needs to be transformed in a system of two first-order equations for 
 the vector :math:`Y = (y, y')`.  It will be convenient to define 
-:math:`\nu = 2 \varepsilon * \omega_0 = c / m` and :math:`om = \omega_0^2 = k/m`::
+:math:`\nu = 2 \varepsilon * \omega_0 = c / m` and :math:`\Omega = \omega_0^2 = k/m`::
 
-    >>> nu_coef = cviscous / mass
-    >>> om_coef = kspring / mass
+    >>> nu_coef = cviscous / mass  # nu
+    >>> om_coef = kspring / mass  # Omega
 
 Thus the function will calculate the velocity and acceleration by::
 
-    >>> def calc_deri(yvec, time, nuc, omc):
-    ...     return (yvec[1], -nuc * yvec[1] - omc * yvec[0])
+    >>> def calc_deri(yvec, time, nu, om):
+    ...     return (yvec[1], -nu * yvec[1] - om * yvec[0])
     ...
     >>> time_vec = np.linspace(0, 10, 100)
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
I am sorry to say, but I wonder if we do not prefer without the math directive. For these specific examples it does not bring a lot, and it increases the chances of compilation failure, or potential contributors not understanding what is going on. I wonder where it falls on simplicity / benefits line? (YAGNI principle).

What do you think?

Naereen:
That's a good question.

#### Already used, in fact
I think the lecture notes already uses ``:math:`` commands from time to time.
See [here for an example](https://github.com/scipy-lectures/scipy-lecture-notes/search?l=rst&q=%3Amath%3A&utf8=%E2%9C%93) showing how much ``.rst`` files already have a ``:math:`` command (13 files for scipy-lectures).

#### For this page
For this page, I added math commands for fractions and differential operators.
I think adding PNG / MathJax math for a pretty df/dx or Omega or nu is worth it.

Some pages already using ``:math:`` use it for very small formula like A or B.
This page on integration uses a lot of greek letters, powers, fractions etc. A non-latex-aware reader might not even know what ``a^b`` means (in Python it is *not* the power..).

#### Conclusion
**TL;DR** yes I think we should use a little bit of ``:math:`` when necessary, and it was the case for this page. Thanks.

Naereen:
adding_math_directive
For example this part clearly wins in clarity with the processed Math equations.

As for chance of compilation failure, both of us checked every added math equation.
And if the compilation fails, a small edit can quickly bring it back, I guess.

GaelVaroquaux:
Agreed for special symbols like 'omega', but not every formula that you changed has special symbols.

Another reason for not relying too much on math is that it induces extra downloads that do add clumsiness on eg mobile devices.

How about we "cut the pear in two" ;) and use math only for formulas where it make special sense, and rely on simple formatting elsewhere?

Naereen:
Alright, so I suggest to merge my PR and then immediately after to "simplify" a few formula.

I agree on the principle about reducing clumsiness on mobile devices.
I guess the lecture notes are usually used on a laptop, not like social networks or news website.
(But you might have access to more precise insight with audience logs of the website maybe?)

GaelVaroquaux:
OK, merging. Thanks a lot.

With regards to mobile devices, I don't have logs, as I didn't want so far to install spying javascripts. The specific mobile devices that I have in mind are tablets. Also, download of small files matter a lot in non-western parts of the world.

Naereen:
True. OK.
Thanks !

PR #222, opened 2016-01-30
Update matplotlib.rst

negm:
There is a typo in the import statement


GaelVaroquaux:
Thanks! Looks like we are not doctesting everything, and hence not capturing all typos

PR #224, opened 2016-03-04
There is a spelling mistake

allxll:


gertingold:
Probably because checks had not passed, this PR, even though valid, was not merged. In the meantime, #240 noting the same problem has been merged so that the typo is corrected. Thanks for your PR nevertheless.

PR #241, opened 2016-08-02
Patch 1

jmcLaughlin-RMS:
Changes variables "l" to "color" and "r" to "rcolor".

The letter "l" looks too much like "1" which confused a student, and presented a teaching opportunity about how not to name variables.  It would probably be better to leave this teaching moment to a different context. 😉


GaelVaroquaux:
Thanks, that's a good change. And the tests are passing.

Merging

PR #242, opened 2016-08-09
Move importing conventions to beginning

brylie:


GaelVaroquaux:
Why?

It seems to me that it is more useful to start the chapter by describing what is numpy, before going into import conventions, which are indeed important, but a bit more of a detail.

brylie:
Sure, a general numpy introduction at the beginning of the file would be useful. It is just nice to have the import convention in place before demonstrating usage with code examples. This can prevent readers from making simple mistakes, by ensuring they are importing the library before calling 'np' or 'numpy'.

GaelVaroquaux:
I am sorry, I feel that it's better not to get into technicalities too early, to be more didactic. So I'd rather keep the way things are.

It's clearly an editorial choice that is debatable, though.

Thanks for the PR, though.

PR #248, opened 2016-08-27
Clean up .travis.yml

lesteve:
Some things come from nilearn and are not needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,10 @@ notifications:
   email: false
 
 env:
-  global:
-    - TEST_RUN_FOLDER="/tmp" # folder where the tests are run from
   matrix:
     # Ubuntu 14.04 versions
     - DISTRIB="conda" PYTHON_VERSION="$TRAVIS_PYTHON_VERSION"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,17 +33,10 @@ before_script:
...
     - make test
 
 after_success:
     # Ignore coveralls failures as the coveralls server is not very reliable
     # but we don't want travis to report a failure in the github UI just
     # because the coverage report failed to be published.
-    # coveralls need to be run from the git checkout
-    # so we need to copy the coverage results from TEST_RUN_FOLDER
-    - if [[ "$COVERAGE" == "true" ]]; then cp "$TEST_RUN_FOLDER/.coverage" .; coveralls || echo "failed"; fi
+    - if [[ "$COVERAGE" == "true" ]]; then coveralls || echo "failed"; fi
```
GaelVaroquaux: Is COVERAGE ever true? I am not sure that we are using coverage with the scipy-lectures.

lesteve: Good point I can remove this as well.

GaelVaroquaux:
That's great. Merging. Thanks a lot.

PR #252, opened 2016-08-27
Conversion to sphinx gallery

gertingold:
This PR contains a partial conversion to Sphinx gallery. It is set up already for some of the example directories but some work remains to be done.

PR #257, opened 2016-08-27
Sphinx gallery

GaelVaroquaux:
Rebased version of @gertingold 's PR #252 : this is to use sphinx-gallery


themes/scipy_lectures/static/nature.css_t
```diff
@@ -601,11 +601,12 @@ div.body img {
...
 
 /* Center figures that are larger than 100% of the figure */
-div.body div.figure {
+/* div.body div.figure {
     margin-left: -200px;
     margin-right: -200px;
     text-align: center;
-}
+  }
+*/
```
GaelVaroquaux: @gertingold : could you tell me where this was in conflict with sphinx-gallery: which specific element of sphinx-gallery it cause to be a problem.

gertingold:
If this is kept in the css file, all images in the gallery are stacked on each other in one line with only a small horizontal shift. Commenting out this part fixes this problem.

GaelVaroquaux:
> If this is kept in the css file, all images in the gallery are stacked on each
> other in one line with only a small horizontal shift. Commenting out this part
> fixes this problem.

OK, so that's in a Gallery. Do you remember where we have a Gallery
visible in the lecture notes? I must confess that I don't.

gertingold:
You will find a gallery for example [here](https://7-1375598-gh.circle-artifacts.com/0//home/ubuntu/scipy-lecture-notes/build/html/intro/matplotlib/auto_examples/index.html).

GaelVaroquaux:
> You will find a gallery for example here.

Great. I will work from there.

If you update your PR fixing the problem that you have found, I will
merge it into mine, as I haven't been able to do you a PR.

gertingold:
Ok, fine, let's proceed like that.

themes/scipy_lectures/static/nature.css_t
Titan-C: I can tell. If you open a gallery webpage. for example:
`intro/matplotlib/auto_examples/index.html`
you will see that all thumbnails are on top of each other because the margin option pull each element by 200px back

conf.py
```diff
@@ -44,10 +38,35 @@
...
+                      'advanced/mathematical_optimization/examples',
+                      'packages/scikit-image/examples',
+                      'packages/statistics/examples'],
+    'gallery_dirs': ['intro/summary-exercises/auto_examples',
+                     'intro/matplotlib/auto_examples',
+                     'advanced/image_processing/auto_examples',
+                     'advanced/mathematical_optimization/auto_examples',
+                     'packages/scikit-image/auto_examples',
+                     'packages/statistics/auto_examples'],
+    'doc_module': 'scipy-lecture-notes',
```
Titan-C: you don't need this, as this lectures are not a python package.

Titan-C:
If I can suggest. I was testing for sphinx-gallery and python3. There are some extensions to sphinx written for this lectures that we might not be needing any more. For example the `plot_directive.py` was last edited 4 years ago, could we move to the current matplotlib plot_directive? It has evolved a lot in the last years, maybe it already solves the problems once forced you to fork from it.

conf.py
gertingold: Do you refer to `doc_module`? I suppose that `examples_dirs`and gallery_dirs` are needed also in the Scipy lecture notes, or am I misunderstanding something?

GaelVaroquaux:
@Titan-C : I am happy getting rid of our own copy of the plot_directive. I guess that it should be done in a separate PR. Though. If you want to go ahead, it would be great.

GaelVaroquaux:
OK, I've fixed the problem with the CSS in my branch.

gertingold:
> OK, I've fixed the problem with the CSS in my branch.

Great. I can confirm that it works fine for the gallery pages.

GaelVaroquaux:
Great. If you fix the rest on your branch, we can merge.

gertingold:
OK. I have to prepare for my trip now, but I should find some time next week.

emmanuelle:
Please also modify the link in this paragraph https://github.com/scipy-lectures/scipy-lecture-notes/blob/master/guide/index.rst#figures-and-code-examples

GaelVaroquaux:
This PR now looks correct: everything seems to be working fine.

But we cannot merge, as the statistics chapter imports seaborn, and this breaks the rendering of the examples. @Titan-C : you know how to solve that. Do you think that you could do a PR in sphinx-gallery to solve it? We will then release sphinx-gallery, and be able to integrate it in the scipy-lectures. This is blocker.

Seaborn's behavior is dirty :pig: :pig2:

GaelVaroquaux:
I restarted a build on this PR, now that the new sphinx-gallery has been released.

GaelVaroquaux:
OK, this should be ready. Could someone review this, maybe by glancing at the generated pages on circleCI.

emmanuelle:
There seems to be a slight problem with the css: on the front page of the lecture notes, in the download frame, icons are not completely in the frame but overlap with its boundary
lectures

conf.py
```diff
@@ -44,10 +38,44 @@
...
+                     ],
+    'doc_module': 'scipy-lecture-notes',
+    'reference_url': {
+        'numpy': 'http://docs.scipy.org/doc/numpy',
+        'scipy': 'http://docs.scipy.org/doc/scipy/reference',
+        'pandas': 'http://pandas.pydata.org/pandas-docs/stable/',
+        'seaborn': 'http://stanford.edu/~mwaskom/software/seaborn/',
+        'matplotlib': 'http://matplotlib.org',
+        'scikit-learn': 'http://scikit-learn.org/stable',
+        'scikit-image': 'http://scikit-image.org/docs/0.12.x/',
```
emmanuelle: http://scikit-image.org/docs/stable

emmanuelle:
LGTM, merging. Thanks!

PR #254, opened 2016-08-27
[WIP] towards python3

Titan-C:
I'm trying to update to python3 keeping backwards compatibility, part of the goal of #243

PR #259, opened 2016-08-28
Add parentheses for print function

michael-hartmann:
This commit adds parentheses for Python's print function. This way the code not
only "works" on Python 2, but also on Python 3.

### DIFF
--- a/advanced/debugging/index_error.py
+++ b/advanced/debugging/index_error.py
@@ -2,7 +2,7 @@
 
 def index_error():
     lst = list('foobar')
-    print lst[len(lst)]
+    print(lst[len(lst)])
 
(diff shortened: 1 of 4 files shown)


GaelVaroquaux:
> This commit adds parentheses for Python's print function. This way the
> code not only "works" on Python 2, but also on Python 3.

Thanks! That's exactly what we need.

@titan-C had started this work (#254), and your PR will be in conflict
with his. I'll wait a bit to see how he is planning to go ahead with his
PR before merging yours.

GaelVaroquaux:
As PR #254 is stalled, I am merging this one. It is a net improvement. Thank you!

PR #260, opened 2016-08-28
Minor changes in contributing instructions

emmanuelle:


guide/index.rst
```diff
@@ -23,34 +23,57 @@ Make sure to read this `Documentation style guide`_ as well as these
 How to contribute ?
 ===================
 
-Choose a topic that is not yet covered and write it up !
+* If you spot typos, unclear or clumsy wording in the lecture notes, please
+  help to improve them. Simple text editing can be done by `editing files
+  in your GitHub fork
+  <https://help.github.com/articles/editing-files-in-your-repository/>`_ of
+  the lecture notes.
```
GaelVaroquaux: Maybe you should above all mention the "edit button", on the top of all pages

guide/index.rst
```diff
@@ -23,34 +23,57 @@ Make sure to read this `Documentation style guide`_ as well as these
...
+
+* Choose a topic that is not yet covered and write it up !
```
GaelVaroquaux: Maybe here you could mention that the right way to do this is first to create an issue on github, so that there is a discussion about the scope of the corresponding chapter.

GaelVaroquaux:
LGTM. Merging.

PR #264, opened 2016-08-29
Fix Travis Python 3 build

lesteve:
Using environment variables, as PYTHON_VERSION="$TRAVIS_PYTHON_VERSION"
does not seem to work.

While I was at it, I removed the virtualenv section which is not needed.

Build from master showing that Python 2.7 is used in the Python 3.4 build:
https://travis-ci.org/scipy-lectures/scipy-lecture-notes/jobs/155856706#L565

Build from my branch showing that Python 3.4 is used in the Python 3.4 build (as it should be):
https://travis-ci.org/lesteve/scipy-lecture-notes/jobs/155877711#L562

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,18 @@
 language: python
 
-python:
-  - "2.7"
-  - "3.4"
-
-virtualenv:
-  system_site_packages: false
-
 notifications:
   email: false
 
 env:
   matrix:
     # Ubuntu 14.04 versions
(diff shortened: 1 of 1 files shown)


lesteve:
ping, python 3 is not tested at the moment which isn't great !

vahtras:
I saw this too - my PR 266 is essentially a duplication of this (I should have read this first) so I am merging this now.

lesteve:
Great thanks!

PR #266, opened 2016-09-08
Another Travis update

vahtras:
This partly revokes some changes I made previously. In the latest commit python2 was installed in both runs (see logs) because of recent adapting to CircleCI and because TRAVIS_PYTHON_VERSION was not expanded as I thought it would. Now the env block in .travis.yml contains to subblocks: *global* for the library versions common for all runs and *matrix* defining two Python versions for two consecutive runs.

The default ubuntu version is 12.04LTS, adding dist:trusty to get a newer OS 14.04LTS (no support for 16.04LTS yet)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ dist: trusty
 sudo: required
 language: python
 
-python:
-  - "2.7"
-  - "3.4"
-
 virtualenv:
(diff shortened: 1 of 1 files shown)


vahtras:
Solved by #264

PR #267, opened 2016-09-08
Reformatting env block in .travis.yml

vahtras:
This blocking of env: in .travis.yml avoids some text duplication


GaelVaroquaux:
Great! Much better. Sorry for the slow review. Merging!

PR #271, opened 2016-10-09
Added a .yml config file for creating a conda env

hrjn:
For a better handling of Python dependencies. Provided that the user has an Anaconda distribution installed, he can get the environment up and running by executing the following lines in a terminal:

```
# Next line is just run once, when "installing" the environment:
conda env create -f environment.yml
# Switch to the custom environment:
source activate scipy-lectures
```

To get back to default environment, run:

```
source deactivate
```

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,15 @@
+name: scipy-lectures
+
+dependencies:
+- python=2.7
+- cython
+- sphinx
+- matplotlib
+- pandas
+- scikit-learn
+- scikit-image
+- seaborn
+- pip
+- pip:
+    - sphinx-gallery
+


GaelVaroquaux:
Sorry for the slow response. I am afraid that scipy-lectures were a bit dormant this school year, as we were all busy with work.

In general I like the idea, and @emmanuelle was mentioning a similar one in #289. Here are two comments though:

- I think that I would prefer having a requirements.txt, which is standard in Python
- We would probably need to define version numbers.

GaelVaroquaux:
Tackled in #302

GaelVaroquaux:
This has been tackled by #302 which generates the environment.yml from a requirements.txt. Thanks for the effort, though

PR #273, opened 2016-10-11
Mayavi example images are flipped

solarjoe:
I flipped the images, structured and unstructed were mixed up.


GaelVaroquaux:
Indeed! Thanks!

PR #274, opened 2016-10-11
Changing random_intergers() to randint()

mhnazeri:
np.random.random_integers() is depricated in numpy 1.11, we better use np.random.randint() instead.

### DIFF
--- a/intro/numpy/operations.rst
+++ b/intro/numpy/operations.rst
@@ -404,7 +404,7 @@ Other reductions
   .. sourcecode:: pycon
 
    >>> t = np.arange(t_max)
-   >>> steps = 2 * np.random.random_integers(0, 1, (n_stories, t_max)) - 1
+   >>> steps = 2 * np.random.randint(0, 1 + 1, (n_stories, t_max)) - 1 # +1 because the high value is exclusive
    >>> np.unique(steps) # Verification: all steps are 1 or -1
    array([-1,  1])
 
--- a/advanced/image_processing/index.rst
+++ b/advanced/image_processing/index.rst
@@ -115,7 +115,7 @@ For large data, use ``np.memmap`` for memory mapping::
 Working on a list of image files ::
 
     >>> for i in range(10):
-    ...     im = np.random.random_integers(0, 255, 10000).reshape((100, 100))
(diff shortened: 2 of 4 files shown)


intro/numpy/operations.rst
```diff
@@ -404,7 +404,7 @@ Other reductions
   .. sourcecode:: pycon
 
    >>> t = np.arange(t_max)
-   >>> steps = 2 * np.random.random_integers(0, 1, (n_stories, t_max)) - 1
+   >>> steps = 2 * np.random.randint(0, 1 + 1, (n_stories, t_max)) - 1 # +1 because the high value is exclusive
```
gertingold: +1, but wouldn't it be better to set the second argument directly to 2?

mhnazeri: I think for clarification it is better to show it by +1 in the first example.

mhnazeri:
I'm going through the tutorial from the beginning and running examples, I will amend this PR as soon as I encounter another randint()

gertingold:
As far as I can see, `random_integers` is present in the following files

- advanced/image_processing/index.rst
- intro/numpy/operations.rst (you have taken care of this one already)
- intro/numpy/array_object.rst
- pyplots/numpy_intro_5.py

mhnazeri:
_Thanks_. I've changed all those files.

GaelVaroquaux:
Thanks for the pull request and the review. Merging

gertingold:
Thanks for your contribution and sorry for looking into this so late. I am still not completely convinced by the explicit +1 in the random walk example because `randint` works like the familiar `range`, so the reader not knowing `random_integers` should not be too surprised. In addition, the correctness of the generated numbers is checked in the next code line. However, in view of #276 it seems appropriate to merge now and possibly amend the change later. 

@GaelVaroquaux: what do you think?

GaelVaroquaux:
@gertingold : I agree with you and just merged.

gertingold:
But what about the +1? Should we leave it as it is or assume that the reader will figure it out if `high` is set to 2?

GaelVaroquaux:
I thought that we could leave it like this.

gertingold:
Ok, fine.

PR #276, opened 2016-11-28
Update array_object.rst

XytHu:
np.random.random_integers has been deprecated, use np.random.randint instead


gertingold:
Thank you for your contribution. This is a partial duplicate of #274. Note also that your replacement is not equivalent because the argument `high` in `randint` is exclusive in contrast to `random_integer`.

PR #277, opened 2016-12-21
Updated Sympy Section

shibasisp:
In reference to the issue #11543 of (Sympy)[https://github.com/sympy/sympy/issues/11543]. I have updated the sympy section according to latest sympy.


gertingold:
It seems that one check (`dsolve` example in line 451) fails because we are asking Travis to use SymPy 0.7.5 (released in February 2014) while I can confirm that the output given in the PR can be obtained with SymPy 1.0 (released in March 2016). May be we should consider to use a more recent version of SymPy in Travis CI?

shibasisp:
Yes! It is obtained with SymPy 1.0. Maybe changing SymPy's version to 1.0 passes the tests.

GaelVaroquaux:
> It is obtained with SymPy 1.0.

I am fine changing the tests on travis. For this, you'll need to edit the
".travis.yml" in the project.

GaelVaroquaux:
You have failing doctests on travis https://travis-ci.org/scipy-lectures/scipy-lecture-notes/jobs/185876551

It's probably just cosmetics, but I fear that we cannot merge like this. Can you have a look and try to fix them please.

Thanks a lot for your hard work.

shibasisp:
It would be a great help if you could tell me how I can run the tests in my local environment. I tried `make test` but it gave me errors. I am beginner to open source and have not dealt with testing before.

gertingold:
In order to develop a solution to this specific problem, it probably makes sense to concentrate on the file in question by running `python -m doctest sympy.rst`.

shibasisp:
Thanks @gertingold ! Also the `dsolve(sin(x)*cos(f(x)) + cos(x)*sin(f(x))*f(x).diff(x), f(x), hint='separable')` part isn't rendering properly when built, although it is working perfectly fine when github or any other online RST viewer renders. I guess it isn't able to wrap properly.

gertingold:
I built the html pages from your modified version and it works fine for me. As far as I can see, the problem is that the line breaks depend on the width allowed by the output device. While such a problem is usually avoided by the  `+NORMALIZE_WHITESPACE` directive, it does not work here because each logical line is actually a multi-line construct. I wonder whether for the purpose of this example, one should fall back to the standard output mode which yields
```
[Eq(f(x), -asin(sqrt(C1/(sin(x)**2 - 1) + 1)) + pi),
 Eq(f(x), asin(sqrt(C1/(sin(x)**2 - 1) + 1)) + pi),
 Eq(f(x), -asin(sqrt(C1/(sin(x)**2 - 1) + 1))),
 Eq(f(x), asin(sqrt(C1/(sin(x)**2 - 1) + 1)))]
```
This output is easier to handle by doctest and not really harder to read than the multi-line output. If the output mode is changed here, a note should probably be added in view of line 121-123 in `sympy.rst`.

shibasisp:
So should I disable pretty printing for `dsolve(sin(x)*cos(f(x)) + cos(x)*sin(f(x))*f(x).diff(x), f(x), hint='separable')` only? and for that do I need to specify anything for the doctest to pass ?

gertingold:
Just for the record: adding to ``init_printing`` the argument ``num_columns=88`` makes the test for ``sympy`` pass. However, this does not really seem like the right way of doing it.

shibasisp:
I tried it, the doc-test passed however it didn't render correctly.Should I make a PR in that case?

GaelVaroquaux:
> adding to init_printing the argument num_columns=88 makes the test for
> sympy pass. However, this does not really seem like the right way of
> doing it.

We can do this in a hidden command, using restructured-text comments.
That way it is executed by the test but not seen by the user.

That would same good to me.

gertingold:
@shibasisp: What is wrong in the rendering (I cannot try it out right now), can it be fixed by adjusting the number of columns? Since Gael agrees with setting the number of columns, trying to tweak things until everything fits is probably worth it.

shibasisp:
@gertingold the rendering didn't get solved. It is as it was earlier. Only the doctest passed. I think It can be fixed by adjusting the number of columns.

shibasisp:
@GaelVaroquaux  I am unclear how that can be done. I don't have any experience with sphinx except this PR.

gertingold:
@shibasisp: The idea of hidden commands mentioned by @GaelVaroquaux is to put a comment containing the appropriate code. As an example, here are two lines taken from ``intro/scipy.rst`` (lines 385/6):
```
.. Comment to make doctest pass
   >>> np.random.seed(42)
```
which sets the seed in order to have a reproducible result.
A comment starts with two dots at the beginning of the line followed by a whitespace and possibly some text as long as it cannot be interpreted in another way, like ``sourcecode`` or so. The comment continues as long as the text is indented. See also the [section Comments in the Sphinx reStructuredText Primer](http://www.sphinx-doc.org/en/1.5.1/rest.html#comments).

shibasisp:
@gertingold thanks for the explanation! If it's not that urgent, I would like to work on this after a week.

gertingold:
@shibasisp Take your time.

GaelVaroquaux:
@shibasisp : any updates? Some people were inquiring about the sympy section.

shibasisp:
If anyone else is interested, I'd like to close my PR.

PR #280, opened 2017-03-13
switch year

gertingold:
The cover of the edition 2016.1 still says that it is the 2015 edition. Even though we are in 2017 now, it might be appropriate to make the cover consistent for the current edition, i.e. replace 2015 by 2016. The pdf has been generated by means of inkscape.


pdebuyl:
We may need a rebuild for other reasons too, see #281

rougier:
I'll modify the cover.

pdebuyl:
@rougier I am ok to push a new version but I'd like to check the protocol:

    git clone https://github.com/scipy-lectures/scipy-lecture-notes
    cd scipy-lecture-notes
    make install

Is that all?

GaelVaroquaux:
The procedure is technically correct, but we usually do a release when we rebuild the website.

I guess that it's not important.

pdebuyl:
I see the benefit of now having too many tags :-) I'll see what can be collected for a bugfix release and come back.

Issue #281, opened 2017-04-12
matplotlib tutorial: examples not showing


amueller:
here:
http://www.scipy-lectures.org/intro/matplotlib/matplotlib.html#other-types-of-plots-examples-and-exercises

![image](https://cloud.githubusercontent.com/assets/449558/24973542/292f2420-1f8d-11e7-8d52-69df776ce9ef.png)

I don't think that's intentional.

pdebuyl:
A "fresh" build does not do this. @GaelVaroquaux what is the procedure for re-building the website?

pdebuyl:
solved by uploading a recent version

PR #284, opened 2017-04-27
use explicit integer division

pdebuyl:
Fixes build with sphinx-gallery

### DIFF
--- a/advanced/image_processing/examples/plot_geom_face.py
+++ b/advanced/image_processing/examples/plot_geom_face.py
@@ -13,7 +13,7 @@ import matplotlib.pyplot as plt
 face = scipy.misc.face(gray=True)
 lx, ly = face.shape
 # Cropping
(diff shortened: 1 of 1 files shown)


pdebuyl:
BTW, I don't know if we have a "do not merge your PRs policy" here

Also, the problem is that the code, when executed via sphinx-gallery, bears the `from __future__ import division` and that the problem was solved similarly in https://github.com/scipy-lectures/scipy-lecture-notes/commit/bda3c3bf6be5eb98270db349bb2a3c4f51bb9eb0

GaelVaroquaux:
> BTW, I don't know if we have a "do not merge your PRs policy" here

No, we're lacking too much developers. It's good to try to get reviews,
in particular if the changes are large: the goal is to avoid a random
walk in the style and content.

pdebuyl:
Ok. The change was small here, so I ended up merging.

I am ok to rebuild the the website (see #280 and #281) can you check the procedure in #280 ?

PR #285, opened 2017-07-25
2.1.2 - Plural inconsistency fix

LFP6:
There was inconsistency as far as usage of singular vs plural ("a function or a class are" -> "functions and classes are", since the rest of the paragraph uses it as plural). Also removed an extra space


gertingold:
Thanks for your contribution.

Issue #289, opened 2017-08-29
move to scikit-image 0.12 for building?


emmanuelle:
Would it be possible to move to scikit-image 0.12 for building the website (in circle.yml)? 0.12 was released in March 2016 and is included in Ubuntu zesty (0.13 in arty). 

I'll take care of the modifications if everybody thinks we do this version bump.

gertingold:
+1 The code examples contain deprecated code which is something we should avoid.

In this context, I am wondering whether we should set up a requirements file. In this way, it would be clear which versions of the packages the lecture notes are intended for and it would be easier to get rid of deprecation warnings and similar errors.

GaelVaroquaux:
> +1 The code examples contain deprecated code which is something we should avoid.

+1.

> In this context, I am wondering whether we should set up a requirements
> file. In this way, it would be clear which versions of the packages the
> lecture notes are intended for and it would be easier to get rid of
> deprecation warnings and similar errors.

+1 Also, we could use this file in the CI configuration.

GaelVaroquaux:
Tackled in #302

pdebuyl:
Changing the deprecated filter `hsobel` to `sobel_h` makes it work with skimage 0.12 and 0.13.

If there is interest, the change is here: https://github.com/pdebuyl/scipy-lecture-notes/tree/fix_hsobel

GaelVaroquaux:
@emmanuelle  @pdebuyl : the build system is now on skimage 0.12, and passing!

A PR to make the corresponding skimage more modern would be welcome.

PR #291, opened 2017-08-29
[MRG] Fix plot_face_tv_denoise.py with skimage >= 0.12

lesteve:
From the [0.12 release notes](https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_0.12.rst#api-changes):

> skimage.filters.denoise_* have moved to skimage.restoration.denoise_*.

### DIFF
--- a/advanced/image_processing/examples/plot_face_tv_denoise.py
+++ b/advanced/image_processing/examples/plot_face_tv_denoise.py
@@ -9,7 +9,11 @@ import numpy as np
 import scipy
 import scipy.misc
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


emmanuelle:
Thanks! We're discusssing (#289) to bump the scikit-image version to 0.12 for building. Then we would only need to import from restoration. I would suggest to accept your PR because it's correct for 0.10 (version used for building now) and it's correct for more recent versions too.

And then I'm planning to work on porting the code to 0.12+ in the next few days.

GaelVaroquaux:
+1 for merging this, and doing a full port later.

PR #295, opened 2017-09-01
Improve make clean

emmanuelle:


GaelVaroquaux:
I cannot merge this because it contains a move of the pyplot to examples with the rst updated.

GaelVaroquaux:
great. merging

Issue #299, opened 2017-09-01
Write a requirements.txt file 


drvinceknight:
Potentially script generating an `environment.yml` file as well.

EDIT: see https://github.com/scipy-lectures/scipy-lecture-notes/blob/master/.travis.yml for reqs.

drvinceknight:
I'm taking this one.

drvinceknight:
Closed by #302

PR #302, opened 2017-09-01
WIP Add requirements.txt and scipt to build env.yml

drvinceknight:
This relates to #299.

Opening a PR to check that it works. Will subsequently trim the
install.sh which will have things that are not needed.


requirements.txt
```diff
@@ -0,0 +1,9 @@
+numpy==1.11
+scipy==0.19
+scikit-learn==0.18
+matplotlib==1.5
+scikit-image==0.12
+sympy==1.0
+statsmodels=0.6
```
gertingold: In the `statsmodels` line there should be two equal signs.

environment.yml
```diff
@@ -0,0 +1,12 @@
+# Do not manually modify this file.
```
GaelVaroquaux: Be more explicit: say that it's generated, and say what generates it.

GaelVaroquaux:
The approach is the good one.

We should probably rename the "continuous_integration" director to "build_utils" (or "build_tools") and move the "requirements-to-environment.py" in there.

Also: minor: I would renamed "requirements-to-environment.py" to "requirements_to_environment.py". The reason being that dashes in Python file names makes them not importable, and hence it is to be avoided.

GaelVaroquaux:
Merging. Great job! Thanks

PR #304, opened 2017-09-01
Shorten ellipsis for the output of np.array?

pdebuyl:
It is very sensitive to the version of numpy, etc, and will fail even when CI
testing is good.


GaelVaroquaux:
LGTM. merging

Issue #297, opened 2017-09-01
Change the Sympy imports to not flatten the namespace


drvinceknight:
Currently at http://www.scipy-lectures.org/packages/sympy.html we are importing `sympy` as follows:

```python
from sympy import *
```

I suggest changing this to:

```python
import sympy as sym
```

This would require modifying various calls in the docs but otherwise making no other changes.

pdebuyl:
Is `sym' a somewhat standard abreviation for sympy? `sympy` is short enough that it could left as is no?

drvinceknight:
Sym is the recommended abbreviation (and also what I personally use in my teaching).

> is short enough that it could left as is no?

Leaving it as `sympy` is an option (just as long as `numpy` but the status quo is `from sympy import *` which I would suggest is not a good approach.

drvinceknight:
I am happy to pick this up and open a PR. If it is preferred to not make the move or to change to anything else that can be closed/modified :)

GaelVaroquaux:
> I am happy to pick this up and open a PR. If it is preferred to not
> make the move or to change to anything else that can be closed/modified
> :)

I have no string opinion on what the best practice is in terms of
shorthand for sympy. Should we simply ask the sympy developers?

drvinceknight:
> Should we simply ask the sympy developers?

Sure I'll ask on twitter.

FWIW in https://books.google.de/books?id=nHDtDAAAQBAJ&pg=PA264&lpg=PA264&dq=sympy+as+sym+best+practice&source=bl&ots=ky-RFj96Rx&sig=qJP8-gx0Ro4F9hVGKaFehHUIXck&hl=en&sa=X&ved=0ahUKEwi9ibyn4IPWAhVC1RoKHfNvD74Q6AEISjAG#v=onepage&q=sympy%20as%20sym%20best%20practice&f=false it has `import sympy as sym`

drvinceknight:
and this tweet (from sympy): https://twitter.com/SymPy/status/678954093914759168

pdebuyl:
fair enough :-)

drvinceknight:
Closed by #307

PR #307, opened 2017-09-01
Unflatten sympy namespace

drvinceknight:
Two things here (happy to revert to simplify the PR):

1. 917a602 implements the discussion at #297: changes `from sympy import *` to `import sympy as sym`.
2. fdfa8ca changes the `init_printing` call: it was called at the start (in a hidden `TODO`) and then called again to show the reader. However this would result in no change of behaviour. This didactically shows the effect of calling `init_printing`.

### DIFF
--- a/packages/sympy.rst
+++ b/packages/sympy.rst
@@ -45,14 +45,14 @@ First Steps with SymPy
 Using SymPy as a calculator
 ---------------------------
 
-SymPy defines three numerical types: Real, Rational and Integer.
+SymPy defines three numerical types: ``Real``, ``Rational`` and ``Integer``.
 
 The Rational class represents a rational number as a pair of two
-Integers: the numerator and the denominator, so Rational(1,2)
-represents 1/2, Rational(5,2) 5/2 and so on::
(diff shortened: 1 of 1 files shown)


packages/sympy.rst
```diff
@@ -45,14 +41,14 @@ First Steps with SymPy
...
 The Rational class represents a rational number as a pair of two
-Integers: the numerator and the denominator, so Rational(1,2)
-represents 1/2, Rational(5,2) 5/2 and so on::
+Integers: the numerator and the denominator, so ``Rational(1,2)``
+represents 1/2, ``Rational(5,2)`` 5/2 and so on::
 
-    >>> from sympy import *
-    >>> a = Rational(1,2)
+    >>> import sympy as sym
+    >>> a = sym.Rational(1,2)
```
GaelVaroquaux: PEP8 tells us that we should have a space after the coma. Let us add it, while we are changing this file.

drvinceknight: 993ebfb fixes the PEP8 throughout.

GaelVaroquaux:
LGTM. Merging. Thank you

PR #309, opened 2017-09-01
Move from sym.solve to sym.solveset.

### LINKED ISSUE #308: Change from `solve` to `solveset` in `sympy`
In http://www.scipy-lectures.org/packages/sympy.html#equation-solving solving equations is shown with `sym.solve`. This has been depreciated in favour of `sym.solveset`. This page has a number of questions and answers describing why. These include:

> solveset has a cleaner input and output interface: 
> solveset returns a set object and a set object takes care of all types of output. 
> For cases where it doesn’t “know” all the solutions a ConditionSet with a partial solution is returned. 
> For input it only takes the equation, the variables to solve for and the optional argument domain over which the equation is to be solved.

My understanding is that at some point `sym.solve` will be removed. As described at http://docs.sympy.org/latest/modules/solvers/solvers.html:

> Note It is recommended to use solveset() to solve univariate equations, sympy.solvers.solveset.linsolve() to solve system of linear equations instead of solve() and sympy.solvers.solveset.nonlinsolve() to solve system of non linear equations since sooner or later the solveset will take over solve either internally or externally.

I am happy to take this on.

drvinceknight:
Closes #308.

This fixes a number of outputs that were not being doctested (the
Jupyter `In` `Out` prompt was being used which is not recognised by the
python doctest module).

Note that I did not change `sym.solve` for the case of the system of
equations (the alternatives are not yet implemented in solveset and I feel
make things more complicated than they need to be). FWIW in my own
teaching I do not consider systems of linear equations using any of the
solvers and instead show students how to use Numpy (or indeed Sympy) to
do this using matrices.

### DIFF
--- a/packages/sympy.rst
+++ b/packages/sympy.rst
@@ -316,7 +316,7 @@ Equation solving
 ================
 
 SymPy is able to solve algebraic equations, in one and several
-variables::
+variables using :func:`~sympy.solveset`::
 
     >>> sym.solveset(x ** 4 - 1, x)
     {-1, 1, -I, I}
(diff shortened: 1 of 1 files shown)


drvinceknight:
5b4a145 fixes those things we discussed @GaelVaroquaux :+1: (using topics).

GaelVaroquaux:
LGTM. Merging

GaelVaroquaux:
Thanks!

Issue #298, opened 2017-09-01
Remove all the .. plot:: directive


GaelVaroquaux:
We need to remove the plot directive (git grep "plot::" shows a dozen) to sphinx-gallery.

Once this is done, we should switch the build system to Py3

gertingold:
I will take a look at that.

GaelVaroquaux:
Fixed by #310!

PR #310, opened 2017-09-01
plot:: directives removed

gertingold:
All ``plot::`` directives have been removed and replaced by the corresponding sphinx-gallery setup (cf. #298).


conf.py
```diff
@@ -50,7 +50,9 @@
 sphinx_gallery_conf = {
     'examples_dirs': ['intro/summary-exercises/examples',
                       'intro/matplotlib/examples',
+                      'intro/numpy/examples',
                       'intro/scipy/examples',
+                      'advanced/advanced_numpy/examples/plots',
```
GaelVaroquaux: Maybe add a comment about why we use a different path here

GaelVaroquaux:
Great!

Thanks so much. Merging!

PR #311, opened 2017-09-01
Change script that in run under python3

michael-hartmann:
* change iteritems to items: This also works with python2, but it needs a bit more memory.
* print is now a function, so add parentheses
* in line 154 and 158 this_costs is an tuple and I think we want to get the first element


GaelVaroquaux:
Great! Thank you

PR #313, opened 2017-09-01
Switch script to new minimize API

michael-hartmann:
The file plot_gradient_descent.py is still missing, though. I will try to do that this evening.


GaelVaroquaux:
LGTM. Merging. Thank you

PR #314, opened 2017-09-01
Fix Fibonacci example and switch to latex

philpep:
In Fibonacci sequence, u_0 is 0, not 1.
Switch to latex for better rendering


GaelVaroquaux:
LGTM. Merging

Issue #270, opened 2016-09-15
Expose each gallery as a "code sample" section, at the end of each chapter


GaelVaroquaux:
For each chapter with an example/auto_example gallery, we should include it at the end of the chapter the corresponding gallery. Examples in the current master are for the matplotlib chapter, or the statistics chapter.

Things to keep in mind are that it should slot in as well as possible in the table of content. For this, we can play with either a toctree to include it (as in the matplotlib chapter), or the include directive (as in the statistics chapter). Also, what we put in the README.txt, used by sphinx-gallery to create the gallery, is important.

GaelVaroquaux:
If people want to pick this up, this is done in the packages/statistics.index.rst file and in the intro/scipy.rst. These files can serve as examples.

gertingold:
I will work on it.

GaelVaroquaux:
Closing as implemented by #316. Thanks @gertingold !

PR #316, opened 2017-09-02
Galleries at end of chapters added

gertingold:
This PR addresses #270. Hopefully, I did not miss a chapter containing figures.


emmanuelle:
LGTM. The numpy gallery is a bit heterogeneous, since a lot of examples do not display images. I don't think it's a big problem though.

GaelVaroquaux:
> The numpy gallery is a bit heterogeneous, since a lot of examples do not display images. I don't think it's a big problem though.

Me neither.

PR #317, opened 2017-09-02
A few modifications in scikit-image and image processing tutorial, mo…

emmanuelle:
…stly

to be consistent with recent changes in scikit-image API.


GaelVaroquaux:
Great! Merging

PR #320, opened 2017-09-05
[MRG+1] Scikit-learn chapter

GaelVaroquaux:
Rewrite fully the scikit-learn chapter:

- More modern idioms in scikit-learn
- More modern scipy-lecture (heavy use of sphinx-gallery)
- More didactic narrative, with more focus on the high-level concepts.

This PR is ready for review and merge, hopefully


rougier:
Would you have a link for reading the output ?(I've problem with building the doc using Python3)

GaelVaroquaux:
> Would you have a link for reading the output ?(I've problem with building the doc using Python3)

You can always look at the circleCI run, and look in the "artifacts" tab.
It's a little hard to find in the circleCI UI, so I am pasting the link
to the last build of this PR below:
https://142-1375598-gh.circle-artifacts.com/0/home/ubuntu/scipy-lecture-notes/build/html/index.html

rougier:
Superficial comments:

Not just this chapter but I think section deep numbering should be removed. For example in this chapter, we end up with "3.6.1.2.2.1. The application problem" which does not bring really useful information (IMHO) and take quite some place. I would prefer to stop at 3.6 and use section title to refer to it (section application problem in chapter 3.6)

Also, the table of content is too narrow. Pretty much every section title take several line. And for some reason, only the first section goes one level down.

rougier:
Still about the look:

Font size between text, block and inline block should be made approximately the same size (it's not the case currently).

GaelVaroquaux:
> Font size between text, block and inline block should be made approximately the same size (it's not the case currently).

Do you mean text and code?

The fact that code in "pre" is larger than text is by design, because it
is more important. Typically, when you teach, you want code blocks to be
well readable.

rougier:
I do not understand this sentence: "Unsupervised learning is used with X, but not y: data without labels. " (section 3.6.6)

GaelVaroquaux:
> I do not understand this sentence: "Unsupervised learning is used with X, but not y: data without labels. " (section 3.6.6)

OK, let me try to formulate it better.

"Unsupervised learning is applied on X without y: data without labels"

Is this better?

GaelVaroquaux:
> Not just this chapter but I think section deep numbering should be removed. For example in this chapter, we end up with "3.6.1.2.2.1. The application problem" which does not bring really useful information (IMHO) and take quite some place. I would prefer to stop at 3.6 and use section title to refer to i> (section application problem in chapter 3.6)

I tend to like explicit numbering, because they facilitate communication
(interestingly, in your next comment, you point to "section 3.6.6").

I would say that stopping at level 2 is really not enough. I would argue
to go at least as far as 4 levels: that's only subsections, given that
we have parts and chapters (to take a latex wording). The good news is
that this is easily feasible with Sphinx. I just tried, and I think that
it does look nice. WDYT?

rougier:
Yes I mean text and code. Even if it is more important, I find it uneasy to read, especially with all the colormaps/boldness around. It tends to hide things. But maybe it's my personal taste. I used it for teaching and it never really bothers me. But reading it on my computer is not really pleasant.

GaelVaroquaux:
> Yes I mean text and code. Even if it is more important, I find it uneasy to
> read, especially with all the colormaps/boldness around. It tends to hide
> things. But maybe it's my personal taste. I used it for teaching and it never
> really bothers me. But reading it on my computer is not really pleasant.

That's interesting, because it's a very small effect. I just looked for
it, and had difficulties seeing it.

rougier:
Ok, some maybe that's related to my machine (osx). I will build it locally and see if there's difference.
By the way, do you conform it can be build with python 3 ? I started to build it from scratch and I had to fix code here and there.

rougier:
"Unsupervised learning is applied on X without y: data without labels"

I just get your point. I started reading the section directly without following the order and it didn't make sense. Now it is a bit clearer but I think it would be worth to have first a more general definition of unsupervised learning and then explain in this case this means etc.

rougier:
hiden -> hidden

rougier:
`Note: Can we trust our results to be actually useful?`

Why is this a note if it is actually a question ?

rougier:
From time to time you use the full name in the text (e.g. sklearn.grid_search.GridSearchCV) and some other time you only use the method name (GridSearchCV). Since they all link back to the sklearn documentation, shorter name should preferred I think (and from the code example, you can also see where they come from).

rougier:
Would be good to have some extra white space before each section (3.6.1) such that when you scroll, you visually see you're entering a new section.

rougier:
"features are most useful" -> "features are the most useful"

rougier:
paramaters -> parameters

rougier:
Iris pictures could be made the same size (or at least the same height)

rougier:
"In a later exercise, we will explore the use of learning curves for the photometric redshift problem."

A link to the exercise would be nice.

rougier:
A Simple Example: the Iris Dataset

Would be good to explain what we're trying to do with this dataset in the introduction of this section (e.g. to name the family a specific iris belongs to, based on its measured features)

GaelVaroquaux:
I don't confirm that it can be built with Python 3. Until recently it was impossible because the plot_directive.py needed a lot of fixes. We got rid of it, but we didn't check.

Happy to take a PR on this. It's something that needs to happen.

Wrt to font size, do you mind posting a screenshot?

GaelVaroquaux:
> From time to time you use the full name in the text (e.g. sklearn.grid_search.GridSearchCV) and some other time you only use the method name (GridSearchCV). Since they all link back to the sklearn documentation, shorter name should preferred I think (and from the code example, you can also see where they come from).

My logic was that I wanted to show where they came from, but only once,
and not everywhere. I think that this is useful: more explicit.

GaelVaroquaux:
> "In a later exercise, we will explore the use of learning curves for the photometric redshift problem."

> A link to the exercise would be nice.

Oops. That exercise was removed. Removing the sentence. Thank you.

GaelVaroquaux:
@rougier : I've addressed your point (aside from the full import path in the names of functions / classes, where I was not sure that I agreed with you).

Thanks a lot for the review!!!

GaelVaroquaux:
For more review, new build available here:
https://144-1375598-gh.circle-artifacts.com/0/home/ubuntu/scipy-lecture-notes/build/html/index.html

rougier:
Here is a screenshot

untitled

rougier:
I'll try again the build using python 3 and make a PR if successful.

rougier:
PR looks good to me. I'll open an issue if I found other problems.

GaelVaroquaux:
@rougier: the font size problem that you saw were related to the fact that you were looking at the page over an https connection. As a result, a CSS was not getting imported (browsers refuse to follow http links when over https). I fixed this, and you can have a look at a new rendering:
https://146-1375598-gh.circle-artifacts.com/0/home/ubuntu/scipy-lecture-notes/build/html/packages/scikit-learn/index.html#validation-curves

rougier:
Yes, much better.

GaelVaroquaux:
Great, merging then!

PR #321, opened 2017-09-05
initiate use of pytest

pdebuyl:
For now, the compiled extensions are omitted.

The docstrings are included with the `--doctest-glob '*.rst'` option.

This PR also relies on having added the file `matplotlibrc`


GaelVaroquaux:
Great! A few comments:

* I think that we should add a comment to the matplotlibrc file to say why it is there, and what it does.

* Is there a risk that the matplotlibrc gets picked up in situations that we don't want it (ie outside of the tests)?

* Why did you keep the tests with nose?

* pytests should be added to the requirements.txt, and not in the .travis.yml file. Also, we should probably remove nose from that file.

pdebuyl:
- I'll add a comment later after the file location issue is certain
- matplotlibrc will get picked by any program started in the `scipy-lecture-notes` directory. Users starting a terminal in that directory would be influenced for instance. I can test the env variable setting and locate the file in `build_tools` or in another location.
- I'll remove the tests with nose. I was happy to have pytest running :-)
- Also, of course, moving pytest to the requirements is more suitable.

Question: do we keep *only* pytest as a test runner? I am fine with this.

GaelVaroquaux:
Locating the matplotlibrc in build_tools sounds like a good choice.

> Question: do we keep only pytest as a test runner? I am fine with this.

I think that that's the right thing to do. Nosetests is slowly coming to
an end of life, and pytest is the recommended choice.

pdebuyl:
As of now, the pytest mode tests all except the program that rely on "to be built" extensions. Both travis and circle-ci are happy, but circle-ci does not call "make test" so that was easy :-)

@GaelVaroquaux your points are all addressed: nosetest is removed, I move and commented in the matplotlibrc file and placed the pytest requirement where it belongs (and update the conda env file with the handy tool).

GaelVaroquaux:
This looks good, but the git history is a bit strange. Let me have a look at what I can do.

GaelVaroquaux:
A rebased version of the PR is available at #327, where it's easier to review.

GaelVaroquaux:
Closing as superseeded by #327, that has been merged. Thanks!

PR #322, opened 2017-09-05
Updated year on PDF cover page

rougier:
* Updated the year on PDF cover
* Renamed old covers with the year in the file
* Added a symbolic link from `cover.pdf` to `cover-2017.pdf` and modified `conf.py` accordingly.


GaelVaroquaux:
Thank you so much.

While we are updating the cover page, I think that it would be nice to add at least Gert-Ludwig Ingold, who hasn't authored a chapter per say, but has been very active in the maintenance lately. And maybe adding an ellipsis ('...') at the end of the list of names would be nice too.

rougier:
I do not find any logic in the author order. Do you have a suggestion on where to  add Gert-Ludwig ?
Or do I reorder everyone alphabetically (but you, Emma and Olaf) ?

GaelVaroquaux:
I don't see the logic either :/

I think that we should either order alphabetically, or order with regards
to the number of commits. But number of commits is a terrible metric, so
I would suggest alphabetical order (probably of last name, it makes more
sense).

rougier:
Done.

GaelVaroquaux:
That's great! And the "many others" is exactly right. Merging! Thank you!

We are making progress towards a new release, and it will be awesome!

PR #324, opened 2017-09-06
Fix for python3

rougier:
This is a small fix to allow for the build using Python3.
Note that this requires Sphinx to be patched (see https://github.com/sphinx-doc/sphinx/pull/4043)


GaelVaroquaux:
I am happy to merge this, but I think that the right fix is actually to
remove plot_directive from the conf.py and to remove the plot_directive
file: we are not using them anymore. It was an oversight to leave them
in.

Do you want to do that, and check that it builds with Python3, or should
I do that?

rougier:
Ok, I'll do it.

GaelVaroquaux:
> Ok, I'll do it.

Thanks!

rougier:
If you just type `make` to build the doc, no image is produced and you got a lof warning such as:
```
WARNING: Could not obtain image size. :scale: option is ignored.
```

It would better to disable the `make` without target or to bind it to a default `make html`

GaelVaroquaux:
> It would better to disable the make without target or to bind it to a default
> make html

Binding it to make html sounds good to me.

rougier:
Got problem with `results = pickle.load(open('helper/compare_optimizers.pkl', "rb"))` in plot_compare_optimizers.py. Not sure how I can read it from Python3 with proper encoding (got complaints from gen_rst.py: TypeError: a bytes-like object is required, not 'str')

GaelVaroquaux:
> Got problem with results = pickle.load(open('helper/compare_optimizers.pkl', in plot_compare_optimizers.py.

That's probably because Python3 fails at loading a Python2 pickle.

I'll try to regenerate this under Python 3.

GaelVaroquaux:
I sent you a PR to your branch which fixes the problem that you are encountering with the pickle.

rougier:
Can I modify travis to test or python3 (it will probably make all subsequent tests slower)

rougier:
Oh, Python 3.5 is already tested. How did it pass previously ?

GaelVaroquaux:
Travis does test under Python 3 (and 2).

What you need is to modify CircleCI to build under 3, as it's the build that you want to modify, and not the test. I think that it would be a good thing, if the build does pass under 3.

GaelVaroquaux:
Oops, I just saw your comment:

travis runs the tests (mostly doctests).

circle runs the build (ie sphinx).

The reason that we use circle to run the build is that it can store the output (aka "build artifacts").

GaelVaroquaux:
I think that this can be merged. Right?

pdebuyl:
I suggest two more changes:

1. Fix the integer division in `advanced/image_processing/index.rst` (line 275)
2. Define the variable `PYTHON` in the makefile and use `$(PYTHON) -m sphinx` as `SPHINXBUILD` to enable `make html PYTHON=python` or `make html PYTHON=python3` (sim. edit for `make test`).

I can make those edits after merging this PR to. The flexibility of using `PYTHON` in the Makefile is not really for the continuous integration where there's only one Python installed but more for users and developers here who try to build and test on their local computer.

GaelVaroquaux:
@pdebuyl : do you want to do these changes and send a PR to master? We can then merge the new PR then.

pdebuyl:
Of course, you can merge this. I'll provide the upcoming PR today.

pdebuyl:
@rougier you mention in the sphinx issue that the problem is fixed. I change line 119 of `conf.py` from `version = p.stdout.read().strip()` to `version = str(p.stdout.read().strip())` or `version = p.stdout.read().strip().decode()` to proceed successfully.

GaelVaroquaux:
Merging! Thank you!

PR #326, opened 2017-09-07
include latex_preamble in latex_elements

pdebuyl:
According to sphinx's documentation, a stand-alone latex_preamble variable is
deprecated since version 0.5 http://www.sphinx-doc.org/en/1.5.6/config.html


GaelVaroquaux:
LGTM. Thanks a lot!

PR #328, opened 2017-09-12
use env PYTHON_VERSION in continuous integration

pdebuyl:
The environment variable PYTHON_VERSION is included in the file
environment.yml at run-time for travis CI.


GaelVaroquaux:
In general, the approach is a great one, thanks a lot!

I see that you have removed the environment.yml from the git.
We had decided to keep it checked in to make it easier for people using conda to work on the scipy-lectures. It would be good to check it back in.

I can see a problem of pinning a Python version outside of CI. Maybe the script that generates it could skip the corresponding line if the PYTHON_VERSION environment variable is not defined. And we would check in the version that does not specify a Python version.

pdebuyl:
1. `environment.yml` is recommited and same as in the master branch.
2. The program `build_tools/requirements_to_environment.py` will pick the environment variable `PYTHON_VERSION` if it exists and include the corresponding line.

GaelVaroquaux:
This is really great, and it works as advetised (ie, travis does run Python 3 tests). Thank you! Merging

pdebuyl:
Yeah :-)

PR #329, opened 2017-09-20
Use double slashes for integer division

pdebuyl:
This is needed for proper building of the examples with sphinx.


GaelVaroquaux:
LGTM. Thanks a lot!

PR #330, opened 2017-09-20
make prompt '>>>' and output unselectable

pdebuyl:
This is done by adding a custom css file that acts on the ".gp" and ".go" classes.


GaelVaroquaux:
Is there a reason that you did not add this to our nature.css? The multiplicity of css files delays loading over http for users.

pdebuyl:
As the `nature.css` file has a `_t` at the end and a copyright notice from the sphinx authors, it seemed as it was a "fixed copy" from somewhere else. I can include the modifications there.

pdebuyl:
ok, 2 force pushes and that should do :-)

I can copy-paste from a few random locations into a terminal!

GaelVaroquaux:
That's very cool! Merging.

PR #225, opened 2016-03-18
Added module initialization code for python 3

rougier:
I've been testing this code during a lesson and the module initialization code is for python 2. I added the code for python3 and make it the default.

Maybe a test from at least another person would be good before merge.

PR #331, opened 2017-09-21
Fix c module python 2 3

pdebuyl:
Make the c module in `advanced/interfacing_with_c/python_c_api/cos_module.c` 2/3 compatible.

Add a note about the filenames of Python 3 extensions in the corresponding text.

Finishes #225 that was only Python 3 compatible.


GaelVaroquaux:
Thanks!

PR #334, opened 2017-10-01
Building instructions for Debian/Ubuntu added

gertingold:
This PR adds a rather explicit description of how to set up a miniconda environment to produce the lecture notes in html and pdf format. In this way, at least users on Linux and MacOS should be able to use the same environment as the one used in the tests. I kept the requirements and Fedora sections even though they seem outdated and might not be needed anymore.


GaelVaroquaux:
This looks great! Merging. Thank you very much.

PR #332, opened 2017-09-28
Python3 build

pdebuyl:
This PR should be merged after @rougier's #324 

- The Makefile is modularized with PYTHON variable
- fix integer division in advanced/image_processing/index.rst
- fix smaller issues in the file conf.py

PR #336, opened 2017-10-01
adapted description to Python 3

gertingold:
After #332 has been merged, the html build works also for Python 3. I adapted the text accordingly.


gertingold:
Should we leave the section on requirements and building on Fedora as they are or should we modify this part before a release?

GaelVaroquaux:
Merging, thanks!

Issue #335, opened 2017-10-01
Releasing soon


GaelVaroquaux:
Just a heads up that I think that we could do a release soon: we have a bunch of major improvements that were done lately.

Things to do:

* [x] Merge #334
* [x] Merge #332
* [x] Maybe switch CircleCI to Python 3 by default
* [x] Check that the PDF renders well

gertingold:
Does the html build work for Python 3? I am encountering a problem which seems to be caused by the fact that `version`defined in `conf.py` is of type `bytes`. A call to `sub` fails in Python 3. That is why I wrote #334 for Python 2. If I am doing something wrong and the build works fine for everybody else, I would rewrite #334 to Python 3 before the release.

pdebuyl:
The line

    version = p.stdout.read().strip().decode()

of `conf.py` succeeds on Python 2.7 and 3.5 on my computer. It is included in #332.

With this PR, I can build the html and run the tests for Python 2 and 3 as:

    make test
    make test PYTHON=python3

    make html
    make html PYTHON=python3

The directories `auto_examples` must be cleaned up after a `make html` at this point if you want to change the Python version however.

gertingold:
Perfect. I did my test without #332, so that's why. I will update #334 to Python 3.

GaelVaroquaux:
> Perfect. I did my test without #332, so that's why. I will update #334 to Python 3.

I am merging #332

pdebuyl:
Cool :-) I'll be glad to see an updated version online.

Maybe a few extra checks can be done before release:
- [ ] Assess #243 
- [ ] Use matplotlib 2.0? Or at least check that matplotlib 2.0 builds the html correctly (I believe it does but I have tons of warnings).
- [ ] quick proofread of the build/contribute instructions? (I can do that).

pdebuyl:
Also, any timeline for the release? 1 week / 2 weeks?

GaelVaroquaux:
I fear that updating to matplotlib 2.0 will be a lot of work, because the implicit dependencies of conda will force us to update more packages, and hence work on the doctests again. I think that this is something that we could do in a later release.

With regards to timeline, I actually wanted to do it last week (as with the school year starting, those lecture notes are really useful, including to me :) ), but I got delayed.

gertingold:
[...]

PR #338, opened 2017-10-02
Improvement of requirement/build information

gertingold:
This PR implements changes as discussed in #335.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -54,17 +54,6 @@ Requirements
 |
 |
 
-.. topic:: Building on Fedora
-
-    As root::
-
-        yum install python make python-matplotlib texlive-pdfjam texlive scipy \ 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -51,20 +49,12 @@ Requirements
 * pandas
 * seaborn
 
+For a complete list of requirements with information on the versions presently used
+to build the SciPy lecture notes, see ``requirements.txt``.
```
GaelVaroquaux: Maybe a `:download:` link, so that whoever is looking at that on the web can download the requirements.txt file?

gertingold: Good point. An alternative would be to use a `literalinclude` of the requirements file. What do you think?

gertingold:
Now, `requirements.txt` is inserted by a `literalinclude`, thus keeping the requirements list always up-to-date.

GaelVaroquaux:
Looks great! Merging. Thank you

PR #343, opened 2017-10-03
Replace Fedora build instructions with generic ones

pdebuyl:
Remove the Fedora build instructions. Nobody in the team currently uses
Fedora and can check the instructions.

Provide generic instructions based on pip instead.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -46,6 +46,9 @@ to build the SciPy lecture notes, can be inferred from the
 .. literalinclude:: requirements.txt
     :start-after: # libraries before the special comment marked
 
+The installation instructions below cover (i) Debian/Ubuntu with miniconda as this is the
+method we use to test and build the material and (ii) generic instructions using pip.
+
 .. topic:: **Building on Debian/Ubuntu**
 
     The environment needed to create an html version of the SciPy lecture notes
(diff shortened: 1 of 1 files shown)


gertingold:
Your recipe fails for me in the step installing scikit-image

    IOError: [Errno 2] No such file or directory: '/tmp/pip-build-CaXNEe/scikit-image/skimage/morphology/_skeletonize_3d_cy.pyx.in'

pdebuyl:
Well, the version of scikit-image that we have pinned for conda has a configuration bug. One can use scikit-image 0.13 instead (this is what I do) but then the conda-mumbo-jumbo and the generic instructions are out of sync. There is probably no way out of it until we update scikit-image :-/

gertingold:
What would break if we were to move to 0.13?

pdebuyl:
I build with 0.13 on my computer. I filed a PR to test the CI with 0.13

gertingold:
The only problem I see with this approach is that if somebody needs a different version of a package for some other project, there will be conflicts.

pdebuyl:
At this point, we should include virtual environments in the instructions. That becomes too much in my opinion. The current situation is quite good: we have a working CI and we can get working instructions for, I suppose, the most common use cases. I admit that I run with globally installed packages and typically I have 3 to 5 docstring errors from the tests and the html builds fine.

PR #344, opened 2017-10-03
update scikit-image to 0.13

pdebuyl:
This PR is to test the CI against scikit-image 0.13


gertingold:
Looks like we can switch to 0.13 and keep the instructions for conda and pip in sync.

PR #347, opened 2017-10-11
change from cm.jet to plt.cm.jet

Atlas7:
The `cm.jet(...)` should be corrected to `plt.cm.jet(...)`. Tested revised script in Jupyter Notebook.


rougier:
In fact, we should any colormap but the jet one.

vahtras:
The suggested change should really be included but I have no idea why circle-ci  fails here.

vahtras:
Reproduced the suggested change in PR363 so the error must have been some previous circle-ci setup. Closing this. Thanks!

PR #351, opened 2017-11-08
scipy.rst -  part Global minimum

bayjan:
I tried it multiple times, but preview shows me that I am doing something wrong in formatting. I hope it can be fixed.

### DIFF
--- a/intro/scipy.rst
+++ b/intro/scipy.rst
@@ -477,10 +477,10 @@ sampling of starting points::
    Try multiple starting points to find the global minimum. 
    For example, in this case:
 
-::
+   ::
 
-   >>> min([optimize.basinhopping(f, x0=x[i]).x for i in range(0,len(x),50)])
-   array([-1.30644001])
+       >>> min([optimize.basinhopping(f, x0=x[i]).x for i in range(0,len(x),50)])
+       array([-1.30644001])
 
 .. note::
 
--- a/intro/scipy.rst
+++ b/intro/scipy.rst
@@ -471,6 +471,17 @@ sampling of starting points::
     packages for global optimization are OpenOpt, IPOPT_, PyGMO_ and
     PyEvolve_.
 
+.. warning::
+
+   :mod:`optimize.basinhopping` may not return the global minimum
+   Try multiple starting points to find the global minimum. 
+   For example, in this case:
+
+::
+
+   >>> min([optimize.basinhopping(f, x0=x[i]).x for i in range(0,len(x),50)])
+   array([-1.30644001])
+
 .. note::
 
    :mod:`scipy` used to contain the routine `anneal`, it has been removed in


pdebuyl:
What do you want to achieve?

If you want to have the example in the box, you must indent it *relative to the warning* as

```
.. warning::

   :mod:`optimize.basinhopping` may not return the global minimum
   Try multiple starting points to find the global minimum. 
   For example, in this case:

   ::

       >>> min([optimize.basinhopping(f, x0=x[i]).x for i in range(0,len(x),50)])
       array([-1.30644001])

```

bayjan:
@pdebuyl Thank you. I changed the text accordingly.

pdebuyl:
Great, now it looks fine :-)

What about showing

    [optimize.minimize(f, x0=x0).x for x0 in linspace(0, 10, 11)]

instead?

Else, the reader does not see that the result is not always the stated minimum. Once this question is resolved, I'll merge the PR.

bayjan:
Well, actually in the scipy-lectures book it was said that opimizie.minimize may find a local minima
> 
> Global minimum: A possible issue with this approach is that, if the function has local minima, the algorithm may find these local minima instead of the global minimum depending on the initial point x0:
```
> >>> res = optimize.minimize(f, x0=3, method="L-BFGS-B")
> >>> res.x
> array([ 3.83746709])
```

and then it was suggested that as a solution one can use `optimize.basinhopping`

> If we don’t know the neighborhood of the global minimum to choose the initial point, we need to resort to costlier global optimization. To find the global minimum, we use scipy.optimize.basinhopping() (added in version 0.12.0 of Scipy).


And only a small reminder/correction I suggested that even `optimize.basinhopping()` may return local minima, so a user is advised to try multiple starting points. This was also described in the documentation of `optimize.basinhopping()`.
_Quoted text are from chapter 5 (part 5.5.2) of the scipy-lectures book._

So maybe more complete suggestion/solution could be:
```
>>> def f(x):
...     return x**2 + 10*np.sin(x)
>>> x = np.arange(-10, 10, 0.1)
>>> min([optimize.basinhopping(f, x0=x[i]).x for i in range(0,len(x),50)])
> array([-1.30644001])

```

pdebuyl:
Hi @bayjan 

I believe that my points complete yours. I wanted to illustrate the intermediate step and show explicitly the fact that basinhoping might also not work. This point is lost when one shows directly the result of calling `min` there.

bayjan:
Hi @pdebuyl . 

Thanks for your help.
I assume it is solved now.

pdebuyl:
@bayjan I am sorry, I think I made myself too critical in my writings. I wanted to know if I could amend your PR, not suppress it.

bayjan:
@pdebuyl No, I don't think it was critical at all. I assumed it as a guideline and you helped in quickly solving this tiny issue, thanks again.

PR #352, opened 2017-11-20
changed colourmap of digits scatter plot to paired

gkiar:


vahtras:
Thanks! Sorry for delayed reply.

PR #354, opened 2017-12-03
Update index.rst

caot:
added plt.show()

### DIFF
--- a/packages/statistics/index.rst
+++ b/packages/statistics/index.rst
@@ -266,6 +266,8 @@ dataframes:
 
     >>> from pandas.tools import plotting
     >>> plotting.scatter_matrix(data[['Weight', 'Height', 'MRI_Count']])   # doctest: +SKIP
+    >>> import matplotlib.pyplot as plt
+    >>> plt.show()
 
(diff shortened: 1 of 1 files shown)


vahtras:
Thanks! I principle you are correct, but there are numerous other places in these lecture notes that a similar change could be motivated leading to a lot of extra text and code repetition. This aspect is well described in the matplotlib introduction. Furthermore, these days it is probably more common to copy code snippets into the Jupyter notebook, so you may not even always want to have the plt.show() in there. I will leave this open, if someone else wants to comment.

pdebuyl:
Should we have a "how to use these lecture notes" section?

GaelVaroquaux:
> Should we have a "how to use these lecture notes" section?

Yes, but I don't think that we should write it now: soon there will be a
new sphinx-gallery release with binder integration, and this will be a
big deal.

GaelVaroquaux:
> Great :-) Let us keep in mind that binder has downtime and that we should
> remain "what if binder is down" friendly :-)

Totally

> BTW, I am not sure I understand the scope: would all docstrings be concerned or
> only the examples that are currently managed by sphinx-gallery?

Only the examples managed by sphinx-gallery for now.

PR #355, opened 2018-01-01
Updated statistics section

tommyod:
I read through and made some updates to the statistics section. Updated some of the code, and rewrote a few sentences.

### DIFF
--- a/packages/statistics/index.rst
+++ b/packages/statistics/index.rst
@@ -424,7 +424,7 @@ A simple linear regression
 Given two set of observations, `x` and `y`, we want to test the
 hypothesis that `y` is a linear function of `x`. In other terms:
 
-    :math:`y = x * coef + intercept + e`
+    :math:`y = x * \textit{coef} + \textit{intercept} + e`
 
 where `e` is observation noise. We will use the `statsmodels
 <http://statsmodels.sourceforge.net/>`_ module to:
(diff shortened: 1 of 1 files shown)


packages/statistics/index.rst
```diff
@@ -850,16 +850,17 @@ Can we conclude that education benefits males more than females?
 
 .. topic:: **Take home messages**
 
-   * Hypothesis testing and p-value give you the **significance** of an
-     effect / difference
+   * Hypothesis testing and the p-value give you the **significance** of an
```
GaelVaroquaux: I don't think that I agree with the addition of the "the". Maybe "p-values" (plural)?

GaelVaroquaux:
Looks great! Thanks a lot. Only one minor comment on the "the".

tommyod:
@GaelVaroquaux Cool! Thanks for the quick feedback. I agree with your comment and have addressed it :+1:

GaelVaroquaux:
Darn, I just realized that your changes from pandas.tools.plotting to pandas.plotting require a recent version of pandas. We have a fixed set of dependencies, to ensure that the code is correct, and continuous integration on them, which failed.

We cannot merge this part of the change. As a matter of fact, we cannot merge if the continuous integration is not happy (as visible in the github PR interface).

tommyod:
@GaelVaroquaux I see. That makes a lot of sense. I should have checked that before making changes. Thanks for the detailed feedback.

I have removed the changes to the code from this PR, but kept the changes to the writing if you want to merge those.

vahtras:
I think it looks good, everything tests ok, no conflicts so I merge this

PR #360, opened 2018-01-14
Replace wrong code for the basic 2d plot

### LINKED ISSUE #358: Wrong code for example 2D arrays (such as images)
http://www.scipy-lectures.org/intro/numpy/array_object.html

In the section 1.3.1.4. Basic visualization, the code for example 2D arrays (such as images) is wrong, also wrong in the pdf version.
  
image

pdebuyl:
The example code was for a 1d plot, I used the code from plot_basic2dplot.py
instead.

Fixes #358

### DIFF
--- a/intro/numpy/array_object.rst
+++ b/intro/numpy/array_object.rst
@@ -432,12 +432,11 @@ Or, if you have enabled interactive plots with ``%matplotlib``:
 
 .. sourcecode:: pycon
 
-  >>> x = np.linspace(0, 3, 20)
-  >>> y = np.linspace(0, 9, 20)
-  >>> plt.plot(x, y)       # line plot    # doctest: +SKIP
-  [<matplotlib.lines.Line2D object at ...>]
-  >>> plt.plot(x, y, 'o')  # dot plot    # doctest: +SKIP
-  [<matplotlib.lines.Line2D object at ...>]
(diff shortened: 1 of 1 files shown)


intro/numpy/array_object.rst
```diff
@@ -432,12 +432,11 @@ Or, if you have enabled interactive plots with ``%matplotlib``:
...
 .. sourcecode:: pycon
 
-  >>> x = np.linspace(0, 3, 20)
-  >>> y = np.linspace(0, 9, 20)
-  >>> plt.plot(x, y)       # line plot    # doctest: +SKIP
-  [<matplotlib.lines.Line2D object at ...>]
-  >>> plt.plot(x, y, 'o')  # dot plot    # doctest: +SKIP
-  [<matplotlib.lines.Line2D object at ...>]
+  >>> image = np.random.rand(30, 30)
+  >>> plt.imshow(image, cmap=plt.cm.hot)    # doctest: +ELLIPSIS
```
rougier: Is there a reason to use specifically the hot colormap here ? (I lack the context)

pdebuyl: I just copied the content of `intro/numpy/examples/plot_basic2dplot.py` as there was a misalignment between the example and the plot. BTW, we don't have a way to include the example code directly - or generating the examples code and plot from the rst?

PR #362, opened 2018-03-15
For travis builds, test on TRAVIS environment var

vahtras:
Purpuse of this section is to deactivate a default virtual environment provided
by travisi

Rather than testing on the absence of a CIRLCE_CI variable, test on TRAVIS
instead which is given for all builds with default value "true"


vahtras:
See also https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

Issue #146, opened 2015-08-28
Add a chapter on working on the scipy-lectures


GaelVaroquaux:
- The chapter should be added last

- It should explain quickly

  - How to build (using sphinx)
  - How to doctest
  - The use of examples to generate figures (ie: figure are not allowed to be checked in)
 - The use of intersphinx to link across projects

- It should stress our style guide:
  - Chapters should be readable / doable as a tutorial in 1.5 hours to 2 hours max (which implies limiting the amount of information)
  - Links (internal, to others projects and to Wikipedia) matter terribly
  - Jargon is to be avoided at all cost
  - The use of restructured text directives, such as topic, note, seealso, tip, sidebar, is strongly encouraged

- It should

emmanuelle:
I did not see this issue before I started working on #260. Some points listed above have been fixed but not all of them.

rootsmusic:
Under "[Objectives and design choices for the lectures](https://github.com/scipy-lectures/scientific-python-lectures/blob/5cbb60368da58b937b0f22fa101072525d7ab96b/CONTRIBUTING.rst#objectives-and-design-choices-for-the-lectures)", `scipy` should be inserted into the following sentence:
> Provide tutorials for a selection of widely-used and stable computational libraries. Currently, we cover pandas, statmodels, seaborn, scikit-image, scikit-learn, and sympy.

PR #367, opened 2018-03-21
Contributing and style

pdebuyl:
See #146


vahtras:
Great, but is now in conflict with  previous merge that has to be resolved first.

pdebuyl:
All conflicts are resolved as far as I (and the GitHub diagnostic) can tell :-)

Issue #377, opened 2018-07-07
Error in section 1.2.2.2. Containers


bugalo:
I might have found an error in section "1.2.2.2. Containers", subsection "Strings". The first square shows the use of single, double and triple quotes, and the second square shows a syntax error. I understand that the second square is not supposed contain this error. 

You might consider to take advantage of this error, and explain here how to use " \' " in strings. If that was the original idea of the second square, then it needs some wording.

gertingold:
Thank you for the remark. I prepared a pull request (#380) adding a brief explanation.

PR #380, opened 2018-07-08
comment added in view of #377

gertingold:
This PR takes care of #377 by adding a comment. In addition the comment lines in the first code example of that section are shifted down by one line to avoid the comment to become part of the string `s`.


gertingold:
The problem flagged by Travis CI is not related to the changes proposed in this PR.

GaelVaroquaux:
Thank you @gertingold !

PR #383, opened 2018-08-14
Optimization chapter: link to asv instead of vbench

jorisvandenbossche:
vbench is no longer maintained, and replaced in usage by asv


pdebuyl:
Hi thanks for the PR. The builds are erroring, which is likely nothing to do with the PR but I'll fix the builds before merging.

Pierre

GaelVaroquaux:
I fixed the CI. I think that we can safe merge.

PR #397, opened 2018-09-03
Updates for matplotlib >= 2

jorisvandenbossche:
Two updates:

- 'spectra' -> 'nipy_spectral' colormap: this should be uncontroversial since the new colormap is already supported since matplotlib 1.3
- 'axisbg' -> 'facecolor' in `add_axes`: the new keyword `facecolor` is only supported since matplotlib 2.0, so for older matplotlib versions (1.5 is used at the moment here) this might be a problem. But from matplotlib 2.2 the 'axisbg' keyword starts to raise an error.

So the main decision is maybe: is it OK to raise the minimum supported version of matplotlib to 2.0


GaelVaroquaux:
> So the main decision is maybe: is it OK to raise the minimum supported version of matplotlib to 2.0

Do it break anything?

jorisvandenbossche:
> Do it break anything?

Hard errors: only the second bullet point I mentioned above (but that is an easy fix if we want to update). 

Then there are of course all the style changes, which might make certain plots look bad/wrong (eg if they were customized for the old style). The coming days I will go through the material a bit and will check if I see any obvious mismatches with the new style.

GaelVaroquaux:
> Then there are of course all the style changes, which might make
> certain plots look bad/wrong (eg if they were customized for the old
> style). The coming days I will go through the material a bit and will
> check if I see any obvious mismatches with the new style.

OK, I'll wait for your feedback to merge.

Thanks!

pdebuyl:
matplotlib 2.2 is still Python 2/3 compatible. We could settle on this.

Does anyone know how hard/easy it would be to have matplotlib 2/3 compatible code? If it is reasonably easy, we should arrange for all the code to move to such a "setting".

pdebuyl:
@GaelVaroquaux @gertingold  @emmanuelle @vahtras  I think we can require matplotlib 2.2 for the build. It makes no sense to keep such an outdated requirement as matplotlib 1.5.

Also, most changes were already applied by another commit! (by Mike Müller in october 2018).

GaelVaroquaux:
> @GaelVaroquaux @gertingold @emmanuelle @vahtras I think we can require
> matplotlib 2.2 for the build. It makes no sense to keep such an outdated
> requirement as matplotlib 1.5.

OK. But it's not _that_ old: only 1 year old. Many systems will still
have a version older.

Thanks for all the work that you are making on scipy-lectures!

PR #400, opened 2018-09-08
CLN: more brackets around print statements

jorisvandenbossche:


GaelVaroquaux:
Wow! Lots of legacy code. A nice reminder that any code that is not tested is broken. 

Merging. Thank you !

PR #401, opened 2018-09-20
make cos_module_np 2/3 compatible

moble:
As in 258d192b, the numpy C-API code needs to be updated to work with both python 2 and 3.  For stylistic reasons, I tried to follow that previous commit as closely as possible — but there are additional changes needed for numpy that make the difference a little more extensive.


pdebuyl:
Dear @moble thank you for the pull request. I thought this issue was solved already but it was not apparently.

PR #404, opened 2018-10-27
Rename `plt.cm.spectral` to `plt.cm.nipy_spectral` 

pya:
The current `environment.yml` lists matplotlib version 1.5 as requirement. This is very old. The current version is 3.0. I tried to build with version 3.0.1 and discovered a few problems. For example `plt.cm.spectral` was renamed to `plt.cm.nipy_spectral`  in version 2.2. I updated all files accordingly.

See: https://github.com/matplotlib/matplotlib/pull/7416


emmanuelle:
Thank you, LGTM!

GaelVaroquaux:
LGTM. Thank you very much @pya . Merging.

By the way: feel free to propose upgrading the environment.yml. It's always a sizable amount of work as things always break :/

pya:
Yes, I am thinking about an approach that takes different versions of libraries into account. Many newbies will likely work with most current version conda or pip installs or that comes with Anaconda. At the same time, other users will need to stick to an older version for various reasons.  Should there be a marker somewhere that indicates with which version this example works for sure? Can this be (semi)-automated somehow? How far should be go here? Are there solutions for a similar problems in other projects we can use or adapt?

GaelVaroquaux:
In my eyes, the big difficulty in supporting versions (not even multiple
ones, just following the changes in packages) is the testing: how do we
make sure that what is written is actually valid. In my experience, it is
hard.

I have no issue bumping a bit the package versions.

PR #406, opened 2018-10-28
Additional changes for PR #404

pya:
PR 404 was incomplete. The color map can also be supplied as a string. 
Therefore, I replaced all strings 'spectral' with 'nipy_spectral' for the keyword argument `cmap`.


GaelVaroquaux:
Good shot! Thank you.

Merging!

PR #407, opened 2018-11-14
1.2.5.4. ‘__main__’ and module loading

prajwel:
this section had a small issue with In [12]. I have removed the line.

### DIFF
--- a/intro/language/reusing_code.rst
+++ b/intro/language/reusing_code.rst
@@ -294,13 +294,11 @@ Importing it:
     In [11]: import demo2
     b
 
-    In [12]: import demo2
-
 Running it:
 
 .. sourcecode:: ipython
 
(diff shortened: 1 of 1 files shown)


pdebuyl:
Hi @prajwel 

Thank you for the feedback! The repeat call to `import` is made on purpose, to show that the module is not reloaded.

Did you have an error while testing the code? If so, could you provide the full traceback?

Regards,

Pierre

prajwel:
Oh, sorry, I thought it was a mistake. I should have tested it before creating the pull request.

pdebuyl:
No worry, we rely on direct testing so feedback is always welcome!

prajwel:
Maybe a newline might make it less confusing after the second import?

PR #408, opened 2018-12-06
Fix Incorrect figure in scikit-learn

### LINKED ISSUE #368: Incorrect figure in scikit-learn
In the scikit-learn chapter you introduce the concept of regularization. You show a comparison of a 4-degree and a 9-degree polynomial fit. Then ask, "which do you prefer? Let's see the ground truth" and repeat the same graph. I suppose you wanted here to show the error for each (which should be higher for the better choice, the 4-degree polynomial).

Link to the section:
http://www.scipy-lectures.org/packages/scikit-learn/index.html#regularization-what-it-is-and-why-it-is-necessary

MLopez-Ibanez:
Fixes #368

### DIFF
--- a/packages/scikit-learn/index.rst
+++ b/packages/scikit-learn/index.rst
@@ -443,7 +443,7 @@ the 9th order one?
 
 Let's look at the ground truth:
 
-.. figure:: auto_examples/images/sphx_glr_plot_polynomial_regression_002.png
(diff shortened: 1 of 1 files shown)


MLopez-Ibanez:
@GaelVaroquaux is this OK?

GaelVaroquaux:
Good catch! Thanks a lot, and sorry for the slow reply. I am merging.

Issue #402, opened 2018-09-20
Memory leak in Numpy C-API example


moble:
Following a [similar a report of a similar issue in my own code](https://github.com/moble/quaternion/issues/92), I found a very snippet of code [here in the scipy lectures](https://www.scipy-lectures.org/advanced/interfacing_with_c/interfacing_with_c.html#numpy-support) that's very close to my own (possibly because I stole mine from here???).  The example shows how to wrap the `cos` function so that it works with numpy arrays.

The problem is that there appears to be an extra `Py_INCREF(out_array)` statement just before the return.  This raises the reference count of `out_array` to 2 (as I've verified by printing out the `Py_REFCNT`).  If I then loop over many calls to `cos_module_np.cos_func_np`, I see the memory usage increase with time.  If I remove that INCREF statement, the memory usage stays constant over time.  My test script looks like this:

```python
import cos_module_np
import numpy as np
import os

print("PID:", os.getpid())

for i in range(10000):
    x = np.linspace(0, 2*np.pi, num=50000);
    y = cos_module_np.cos_func_np(x)
    z = x+y
```

I've tested it with a current version of numpy and both python 2.7 and 3.6.

*However*, I'm not making this into a PR because I'm not entirely sure if that's the correct solution.  When I do the analogous thing with my own code, I eventually get segfaults after ~100 iterations.  I don't see such segfaults with this code, but our codes really look the same fundamentally, so I'm wondering if this is the correct approach.

seberg:
Random comment, I find that snippet a bit strange, because NpyIter has nice stuff to iterate both arrays at once, it will even allocate the out array for you if you want. Other then that, I do not see why the INCREF there is needed, I do think that `out_array` should be initialized to NULL probably to make the error path save. Anyway, would have to carefully check things as well to be sure.

moble:
@seberg Would you say that the modern approach would basically follow [this example](https://www.numpy.org/devdocs/reference/c-api.iterator.html#simple-multi-iteration-example) from the numpy C-API docs?  The main changes I see would be (1) some obvious arg-handling stuff and (2) replacing
```C
memcpy(dst, src, itemsize);
```
in the inner loop with something more like
```C
*(double *) dst = cos(*(double *) src);
```
Does that sound right?

seberg:
[...]

PR #411, opened 2019-02-08
ENH: Make the numpy advanced iterator example more elaborate

seberg:
This uses the iterator to allocate the output. It also adds some
checks so that the code will not fail on non-native byte order
or unalinged inputs.
It also makes use of the inner loop optimization which should be
used for most iterations such as here.

Possibly this goes too far and a simpler version is better?

Closes gh-402.

---

It should probably be good, but I should go over it one more time myself before merging. I am not sure if this may be adding a bit too many bells and whistles for a simple tutorial.


pdebuyl:
The past example was not "hello world" neither, so I see no issue with the level of complexity.

I had a look and tested the code with Python 2.7, 3.6 and 3.7.

Did you want to check this further or can I merge?

seberg:
@pdebuyl went through it, just changed the comments/whitespace a bit. I think it should be OK, the question is if it is didactically nice.

I added more involved tests for the corner cases to the test file, not sure if this is something you like or if it would have the right style.

pdebuyl:
Hi @seberg 

Sorry for the (even for me) unusually long delay. As I am not familar with the NumPy C API it took me a few times to get to the code. A lot of the boilerplate is standard, which is good :-)

I have one issue though. The numerical results are not correct. I printed y for completeness:

```
[ 1.          0.1015857   0.1015857  -0.16299078  0.1015857   1.
 -0.16299078 -0.78223089  0.1015857   0.7421542   1.          0.1015857
 -0.16299078  0.7421542  -0.78223089  1.          0.1015857  -0.16299078
  0.7421542  -0.78223089  1.          0.7421542   0.1015857  -0.78223089
 -0.16299078  1.          0.7421542   0.1015857  -0.78223089 -0.16299078
  1.          0.7421542   0.1015857  -0.78223089 -0.16299078  1.          0.7421542
  0.1015857  -0.78223089 -0.16299078  1.         -0.78223089  0.7421542
  0.7421542   0.1015857   1.         -0.78223089  0.7421542  -0.16299078
  0.1015857   1.         -0.78223089  0.7421542  -0.16299078  0.1015857   1.
 -0.78223089  0.7421542  -0.16299078  0.1015857   1.         -0.78223089
  0.7421542 ]
```

I can't find where the error comes from. I could reproduce on two linux computers (ubuntu 16.04 and debian sid), Python 2.7, 3.6 and 3.7.

Looking at the documentation, I can't see what is wrong :-(

seberg:
Thanks for having a look. I added an `assert()` for the out stride (that using `out++` is indeed fine).

Other than that, the issue was that I wrote `(double) *in` instead of `*(double *)in`...

pdebuyl:
Thank you very much for the fix!

PR #413, opened 2019-02-20
Include Visual Studio Code to intro

Hofer-Julian:
It offers excellent python and anaconda support and I think it should be included to the list


GaelVaroquaux:
LGTM, +1 for merge when CI has ran.

pdebuyl:
What about linking to the Python language page directly?

https://code.visualstudio.com/docs/languages/python

Hofer-Julian:
@pdebuyl True, that's better

GaelVaroquaux:
Thanks. Merging!

PR #414, opened 2019-02-20
More detailed install instructions

Hofer-Julian:
I remember that when I learned python with scipy-lectures, I felt left alone with this part.
I knew how to install packages on my Ubuntu machine, but I didn't know which packages I need and how they are called on my distribution. I guess one could argue about the phrasing, but I would appreciate it if there would be any form of install instructions for python-beginners (and in the ideal case even for linux-beginners).


pdebuyl:
Thank you!

PR #416, opened 2019-03-17
Corrected link to Dive into Python

epogrebnyak:
Existing link http://www.diveintopython.net/ no longer pointed to the proper website.

### DIFF
--- a/intro/language/python_language.rst
+++ b/intro/language/python_language.rst
@@ -12,7 +12,8 @@ Gaël Varoquaux*
     necessary for getting started with Numpy and Scipy is addressed here.
     To learn more about the language, consider going through the
     excellent tutorial https://docs.python.org/tutorial. Dedicated books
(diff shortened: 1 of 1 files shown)


gertingold:
Thanks for spotting this problem. May I suggest to adapt the title of the link? The book is now called "Dive into Python 3".

epogrebnyak:
Sure, that would be a proper title. You can also extend the book list, if you have space for it. Some suggestions [here](https://github.com/epogrebnyak/learn/blob/master/pure_python_guide.md#introductions).  I think [The Hitchhiker’s Guide to Python](https://docs.python-guide.org) is a particularly vibrant and practical title too.

gertingold:
Would you mind adding a commit with the updated title to this PR to make it complete?

gertingold:
Thanks again for your contribution.

PR #420, opened 2019-04-24
fix typo: indivuals -> individuals

staeiou:
Thanks for the tutorial!


pdebuyl:
Thank you for the fix!

PR #421, opened 2019-04-30
Update python_language.rst

nicola-masarone:
Previous link to: 
http://www.diveintopython.net/ 
doesn't work anymore.


pdebuyl:
Hi, thank you for contributing!

The existing link works fine as far as I can see. Unless others confirm the error, I will let the URL stand as is.

gertingold:
It seems that @nicola-masarone has a point. I also get a message that the domain is for sale.

pdebuyl:
Ok, it was already fixed in 88172b1f004f23aed3302bf028b2615b45cfbce5

We should update :-)

PR #426, opened 2019-05-01
Update exceptions.rst

nicola-masarone:
To better explain what pass statement does: nothing :-)


GaelVaroquaux:
Thanks!! Merging.

PR #442, opened 2019-06-18
Fix math error

andreas-h:
I believe in the damped oscillator example, there's a factor $2$ missing in the calculation of the derivative vector.


gertingold:
Thanks for spotting this mistake.

PR #449, opened 2019-08-23
small text update in broadcasting example

sjvrijn:
distance to origin example is shown for 5x5 grid, but text said 10x10


gertingold:
Thanks for noticing and correcting this.

PR #451, opened 2019-09-06
remove endpoint=True from np.linspace()

timhoffm:
There is no need to specify this explicitly because it is already the
default behavior.


GaelVaroquaux:
Indeed, it is no longer needed. Thank you! Merging.

PR #467, opened 2020-01-17
Update packages

pdebuyl:
Update of packages, and of Python version to 3.6.

- Update pytest, add pytest.ini file with options NUMBER and NORMALIZE_WHITESPACE
- Update of NumPy
- Update of other packages to find a compatible mix.


pdebuyl:
Thank you for the review @gertingold !

Regarding the "+SKIP": depending on the machine, the result is of the type "0... e-17" or "-0... e-17". The change in sign modifies the spacing in the printing of the array and that breaks the test. I did the test on my computer with the same version of Python and of all the packages in a conda environment, which should be the same as for the CI machine, but it was impossible to format the test in a robust way :-(

The problem here, in a way, is that the doctest method is too stringent for numerical results. A numpy "assert_allclose" would do for the values, but not for the representation.

From the creation of the notes, the tests have been made very robust with respect to minor issues but this is beyond what we can do. We would need a doctest extension for numerical work...

I asked for two reviewers in the hope of getting at least one :-) So I'll merge soon if you are ok with the skipped test.

gertingold:
Thanks for the explanation, @pdebuyl. I was sure that you had a good reason to skip it, but it is good to know the motivation. From my side, you can go ahead and merge.

pdebuyl:
Thanks :-)

vahtras:
Excellent!

PR #473, opened 2020-01-23
replace suggestions for debugging environments

### LINKED ISSUE #453: Broken link to winpdb
The link to http://winpdb.org is broken.

Winpdb does not seem to be maintained anymore. The latest version is from 2010, see https://pypi.org/project/winpdb/#history. There is a reborn version at https://github.com/bluebird75/winpdb, see also https://pypi.org/project/winpdb-reborn/. For the moment, there seems to be only a stable version for Python 2, while the version for Python 3 is under development.

Can we recommend the use of `winpdb` at this point?

pdebuyl:
fixes #453

### DIFF
--- a/advanced/debugging/index.rst
+++ b/advanced/debugging/index.rst
@@ -473,15 +473,14 @@ Other ways of starting a debugger
 
 .. topic:: Graphical debuggers and alternatives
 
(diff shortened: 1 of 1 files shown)


pdebuyl:
thanks @vahtras

Issue #262, opened 2016-08-28
Replace old figure with generated one


emmanuelle:
Figure http://www.scipy-lectures.org/_images/random_c.jpg is not so nice and a static file (disclaimer: it's an old figure from my PhD work =p). It would be better to replace it with a cool figure generated from a Python script.

GaelVaroquaux:
Maybe bessel functions, as in the following Chaco plot:
http://docs.enthought.com/chaco/user_manual/annotated_examples.html#bigdata-py

gertingold:
It would be nice to have figures for which the relation to the packages is more obvious. Following the suggestion of displaying Bessel functions, such a figure would nicely relate to the SciPy package, while a second image of a (random?) matrix using `imshow` could be appropriate for NumPy.

lawrencefchan:
Has this been implemented anywhere in the code? I'm looking for examples to create a consistent style, but I haven't been able to find any.

In the example given in [how to contribute](http://scipy-lectures.org/guide/index.html#figures-and-code-examples) it doesn't look like there is a reference to `plot_xxx.py`.

pdebuyl:
Thank you for the comment. The practice has been used to progressively replace the content but we have not documented it yet.

For the example you are working on, you can look at the file stats-interpolate.rst in the same directory, it uses the updated technique.

In short:

1. create a directory named "examples" (already done for the directory here).
2. Add a README.rst (necessary for sphinx-gallery)
3. add a file `plot_XXX.py` with a top "docstring" (a short description between triple quotes)
4. write the code to generate a figure

the figure will be named "auto_examples/images/sphx_glr_plot_XXX_001.png" (numbering from 001 onwards if several figures).

pdebuyl:
Proposal: add fives terms for a Fourier series. Quite simple and visual.

```
import numpy as np
import matplotlib.pyplot as plt

x = np.linspace(0, 2, 201)
terms = [-np.sin(k*2*np.pi*x)*(-1)**k for k in [1, 2, 3, 4, 5]]
series = np.sum(terms, axis=0)
[plt.plot(x, terms[i], alpha=1/(i+1)) for i in range(len(terms))]
plt.plot(x, series)

plt.show()
```

PR #474, opened 2020-02-09
Added graphical solution to example

lawrencefchan:
Added the fitted curve to demonstrate optimize-fit example.

Also added some lines to the plot code snippet to reflect output graph more closely.


intro/summary-exercises/optimize-fit.rst
```diff
@@ -56,8 +56,14 @@ and visualize it::
...
     >>> t = np.arange(len(waveform_1))
-    >>> plt.plot(t, waveform_1) #doctest: +ELLIPSIS
+
+    >>> fig, ax = plt.subplots(figsize=(8, 6))
+    >>> ax.plot(t, waveform_1) #doctest: +ELLIPSIS
     [<matplotlib.lines.Line2D object at ...>]
+    >>> ax.set_xlabel('Time [ns]') #doctest: +ELLIPSIS
+    Text(0.5,0,'Time [ns]')
+    >>> ax.set_ylabel('Intensity [bins]') #doctest: +ELLIPSIS
+    Text(0,0.5,'Intensity [bins]')
```
lawrencefchan: I wasn't sure whether `#doctest ELLIPSIS ` should have been included, but my code didn't pass the tests without it.

pdebuyl:
Dear @lawrencefchan 

Thank you for the contribution. I will review it later this week.

pdebuyl:
Hello, here are my comments on the pull request:

1. There are existing graphics file, which we try to remove (see #262). For any update to the project, we do not take figure files but request that figures are generated: http://scipy-lectures.org/guide/index.html#figures-and-code-examples

    Could you transform your pull request to autogenerate the figures? Also, the third figure is superfluous.

2. The `+ELLIPSIS` option is only necessary when the code returns the object id, which is the case for plots and legends but not for the labels where they can be removed.

3. The usage of `subplot` and the related `ax.` method calls introduce unnecessary complexity. Please use the simpler `plt.plot`, `plt.xlabel`, etc.

I realize that these are big changes to your pull request, please let me know if you are interested to go on with the modifications.

Pierre

lawrencefchan:
@pdebuyl thanks for the comments. I'll fix it and try again. Probably this weekend.

lawrencefchan:
@pdebuyl sorry it's taken so long. how do the most recent changes look?

pdebuyl:
I'll give it a look soon. Sorry for the delay. The merges end up with a lot of changes, many unrelated to your proposition. I'll have to sort it out.

lawrencefchan:
Would it be easier if I closed this pull request and opened a new one from the current master?

pdebuyl:
Yes, a pull request with only your changes would be easier.

Also:

1. The line "Solution: :ref:`Python source file <sphx_glr_intro_summary-exercises_auto_examples_plot_optimize_lidar_data.py>`" should be de-indented (else it is part of the "image" directive that takes no comment". Same for "Compare the result of :func:`scipy.optimize.leastsq`".

2. I'd like to keep the current narrative version, as in the rest of the lecture notes, where the code is show below its explanation. "load the waveform" followed by a few lines of Python code, and so on. There would be some duplication of code between the lecture page and the sphinx-gallery generated page from the Python file but that seems the best solution to me.

3. For many matplotlib commands, you don't need `#doctest: +ELLIPSIS`. This is only necessary for those that return a memory address ("line object at ..." for instance).

PR #475, opened 2020-03-01
add python 2/3 section in intro

### LINKED ISSUE #463: Add section on Python 2 / 3
Python 2 becoming unsupported, I propose that we add a section on the whole Python 2 / 3 issue.

We should not patronize users. I propose to remain factual, pragmatic, and helpful.

- What is the transition from Python 2 to Python 3?
- How does it impact the usability of Python for science?
- Agenda of the migration: many packages are dropping support for Python 2 in coming releases.

We should probably release a version (or all 2020 releases if several) of the notes with Python 2/3 compatibility and  let it be available by some means as a maintenance branch (only bug fixes).

pdebuyl:
fixed #463

### DIFF
--- a/intro/index.rst
+++ b/intro/index.rst
@@ -18,6 +18,7 @@ itself, to numerical computing or plotting.
 
    intro.rst
    language/python_language.rst
+   python_2_python_3.rst
    numpy/index.rst 
    matplotlib/index.rst 
    scipy.rst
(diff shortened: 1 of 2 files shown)


pdebuyl:
@scipy-lectures/developers @vahtras I would like feedback on this new section on Python 2/3

pdebuyl:
Thanks!

PR #485, opened 2020-06-10
Update functions.rst

BSGalvan:
Changes made:
1. Line 292 - From Python 3.6 onwards, the order of the keyword arguments in the dictionary returned preserves the order of the keyword arguments as they were input in the function call (as recommended in https://www.python.org/dev/peps/pep-0468/).  The output on Line 292 hence is edited to reflect this.
2. Line 351 - Same as above.


pdebuyl:
Thank you for the contribution!

The code you fixed is in "ipython source blocks" and is not tested automatically. We use Python 3.6, so this should have been catched.

PR #489, opened 2020-07-09
Added graphical solution to optimize-fit

lawrencefchan:
A graphical solution to optimize-fit has been added.
Additionally, file structure has been organised to be more consistent with the other summary exercises (e.g. image-processing) as follows:
- Python code for generating charts has been included under examples
- The .npy binary files have also been moved under examples

This pulled from the most recent master to simplify the merge process.


pdebuyl:
Thank you @lawrencefchan for your contribution!

I did remove the last figure (which is the solution to the advanced exercise) and use image targets for the source codes.

PR #498, opened 2021-01-10
Fix inner plot in fftpack example

OsaydAbdu:
The inner plot incorrectly has the peak at 0.25 where the peak frequency should be at 0.2.

Explanation of fix: 
The x axis plots the `freqs` variable after removing the non-positive values while the y axis plots `power` without removing the non-positive values and that's what causing the error. The fist just removes the non-positive values in the y axis as well.


pdebuyl:
Thank you for this fix!

pdebuyl:
PS: the website is not in sync with the repository, the fix will appear at the next release.

PR #499, opened 2021-11-17
Update links in readme

zmoon:
It might be cool if <scipy-lectures.org> would default to HTTPS as well.

Also it seems like there is a more recent Zenodo release than the one currently linked.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@ ecosystem that can be used for a full course of scientific computing with
 Python.
 
 These documents are written with the rest markup language (``.rst``
-extension) and built using Sphinx: https://www.sphinx-doc.org.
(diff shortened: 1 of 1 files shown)


pdebuyl:
Hi @zmoon thanks for submitting the PR. I'll check it tomorrow.

Regarding the DOI, we use the generic zenodo DOI that points to the latest version:

> You can cite all versions by using the DOI 10.5281/zenodo.594102. This DOI represents all versions, and will always resolve to the latest one

I see thanks to your comment that the 2020 release appears in beta and is not picked up, which means that I might need to re-release it.

zmoon:
> This DOI represents all versions, and will always resolve to the latest one

I see that now. It's interesting that the more recent one, because marked as pre-release, is treated as older on Zenodo.
image
("newer version" takes you to the 2017 one)

PR #511, opened 2022-09-13
Remove Python 2 material 

jarrodmillman:


pdebuyl:
Here, you remove the reference to http://scipy-lectures.org/intro/python_2_python_3.html but keep the file, is it intentional?

jarrodmillman:
I think I removed it in the first commit: f5dd75531cb

Am I missing something?

PR #513, opened 2022-09-13
Rename master to main

jarrodmillman:
Here is what I think needs to be done:

- [x] Rename default branch `main` on GitHub:
    - https://github.com/ipython/ipython/settings/branches
- [ ] Merge this PR
- [ ] Update your locally cloned repository to have a `main` branch as follows:
```
git branch -m master main
git fetch <YOUR_UPSTREAM_REMOTE>
git branch -u <YOUR_UPSTREAM_REMOTE>/main main

(where YOUR_UPSTREAM_REMOTE is typically called `upstream` or `origin`)
```

The whole process should only take a few minutes.

PR #515, opened 2022-09-13
Use pip on circleci

jarrodmillman:
@pdebuyl @GaelVaroquaux  This is ready to merge. I was getting frustrated trying to sort out what versions of what packages were available on conda-forge (e.g., the current failures on master are due to missing packages on PyPI). Everything is on PyPI and it works without causing any issues. It also removes old custom code and replaces it with standard commands.

Once we get everything updated and tested, I can add back tests for conda. But I would prefer to do the updates without the hassle of sorting out conda issues.


jarrodmillman:
Once this is merged, I will rebase my other PRs on master and they should be ready to merge.

pdebuyl:
So, I should do this one before #508 #510 and #513 is that right?

PR #528, opened 2022-09-27
Use solve ivp

### LINKED ISSUE #483: Recommend solve_ivp instead of odeint in section 6.7.2
Hello!

I was reading the section 6.7.2 on numerical integration of differential equations and I feel like it might need some updating. The recommended way of doing it is to use [scipy.integrate.odeint](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.odeint.html#scipy.integrate.odeint), but checking its documentation it recommends using [scipy.integrate.solve_ivp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html#scipy.integrate.solve_ivp) instead.

I just came across this lecture notes yesterday, so I wouldn't know what kind of fix would be best. Whether it's rewriting the section with the new function, keeping both options or just referencing solve_ivp as one of those "Tips", I know not.

pdebuyl:
Fixes #483

### DIFF
--- a/intro/scipy.rst
+++ b/intro/scipy.rst
@@ -784,28 +784,33 @@ Integrating differential equations
 :mod:`scipy.integrate` also features routines for integrating `Ordinary
 Differential Equations (ODE)
 <https://en.wikipedia.org/wiki/Ordinary_differential_equation>`__. In
-particular, :func:`scipy.integrate.odeint` solves ODE of the form::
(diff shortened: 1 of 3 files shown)


jarrodmillman (approved):

pdebuyl:
Thanks for looking at it

PR #547, opened 2022-11-03
Make code block self-contained

habi:
I had to add this, to make the block run by itself.


stefanv:
Good catch, thanks. In the future, we'll probably transition to the `import skimage as ski` style.

PR #559, opened 2022-12-02
Pin to sphinx 5.2.3

jarrodmillman:
TODO: math roles aren't rendered with sphinx 5.3


jarrodmillman:
With sphinx 5.3, we get
screenshot

For example, with sphinx 5.3, in the left column we have
```
-1 < x_1 < 1
```

PR #575, opened 2023-05-22
Fix casing for SciPy and NumPy

tupui:


jarrodmillman (approved):

jarrodmillman:
@tupui Can you take a quick look at the failure?

mdhaber (approved):
I'm not certain that this is everything, but something is better than nothing, and I didn't find an incorrect change.

intro/numpy/array_object.rst
```diff
@@ -107,7 +107,7 @@ NumPy Reference documentation
    .. sourcecode:: pycon
 
      >>> help(np.array) # doctest: +ELLIPSIS
-     Help on built-in function array in module numpy:
+     Help on built-in function array in module NumPy:
```
mdhaber, line 110:
```suggestion
     Help on built-in function array in module numpy:
```

PR #577, opened 2023-05-23
Update usages of np.random to use a generator

tupui:


tupui:
Hopefully now it's all. I suggest leaving the update of doctests for a follow up. This is already quite large.

stefanv (commented):
This looks great, thanks @tupui! I'd get rid of the long seeds (of all the seeds!), but we can leave that for another PR as well.

stefanv (approved):

rossbar (approved):

jarrodmillman (approved):

PR #580, opened 2023-05-23
MAINT: remove duplicate eigenfaces header

mdhaber:
Closes gh-560

I removed the header from the index rather than `plot_eigenfaces.py` in case `plot_eigenfaces.py` is accessed as a stand-alone document. Need to check with the sidebar thing works properly.

### DIFF
--- a/packages/scikit-learn/examples/plot_eigenfaces.py
+++ b/packages/scikit-learn/examples/plot_eigenfaces.py
@@ -2,13 +2,6 @@
 The eigenfaces example: chaining PCA and SVMs
 =============================================
 
-
-.. sidebar:: Code and notebook
-
-   Python code and Jupyter notebook for this section are found
-   :ref:`here
-   <sphx_glr_packages_scikit-learn_auto_examples_plot_eigenfaces.py>`
-
 The goal of this example is to show how an unsupervised method and a
 supervised one can be chained for better prediction. It starts with a
 didactic but lengthy way of doing things, and finishes with the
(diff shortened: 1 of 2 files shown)


mdhaber (commented):
I don't think this should be merged as-is because it doesn't seem like the links are working as they're supposed to.

Instead, I wonder if this should even be a section of `index.rst`?  It is an example, so it appears with the other examples at the end of the page. Why don't we just remove it from here?

packages/scikit-learn/examples/plot_eigenfaces.py
```diff
@@ -2,6 +2,13 @@
 The eigenfaces example: chaining PCA and SVMs
 =============================================
 
+
+.. sidebar:: Code and notebook
+
+   Python code and Jupyter notebook for this section are found
+   :ref:`here
+   <sphx_glr_packages_scikit-learn_auto_examples_plot_eigenfaces.py>`
+
```
mdhaber, line 11:
```suggestion
```

packages/scikit-learn/index.rst
```diff
@@ -1358,14 +1358,8 @@ of digits eventhough it had no access to the class information.
...
-.. sidebar:: Code and notebook
-
-   Python code and Jupyter notebook for this section are found
-   :ref:`here
-   <sphx_glr_packages_scikit-learn_auto_examples_plot_eigenfaces.py>`
+.. The eigenfaces example: chaining PCA and SVMs
+.. =============================================
 
 
 .. include:: auto_examples/plot_eigenfaces.rst
```
mdhaber, line 1365:
```suggestion
```
(We would delete the remaining lines, too.)

mdhaber (commented):
I don't think this should be merged as-is because it doesn't seem like the links are working as they're supposed to.

Instead, I wonder if this should even be a section of index.rst? It is an example, so it appears with the other examples at the end of the page. Why don't we just remove it from here?

jarrodmillman (approved):

PR #583, opened 2023-05-23
ENH: update scipy.optimize

mdhaber:


intro/scipy/index.rst
```diff
@@ -278,18 +278,70 @@ interpolation example.
...
+
+    >>> def f(x):
+    ...     return (x-1)*(x-2)
+    >>> res = sp.optimize.root_scalar(f, x0=0)  # x0 is the guess
+    >>> res
+      converged: True
+           flag: converged
+ function_calls: 12
+     iterations: 6
+           root: [ 1.000e+00]
```
mdhaber, line 301:
```suggestion
         converged: True
              flag: converged
    function_calls: 12
        iterations: 6
              root: [ 1.000e+00]
```

intro/scipy/index.rst
```diff
@@ -278,18 +278,70 @@ interpolation example.
...
+
+Note that only one the root at ``1.0`` is found. By inspection, we can tell
+that there is a second root at ``2.0``. We can direct the function toward a
+particular root by changing the guess or by passing a bracket that contains
```
mdhaber, line 305: I would like to show that changing the guess to a point where the derivative is 0 can cause problems with the gradient-based methods. Would that be too much?

mdhaber (commented):
Edits

mdhaber:
Test is failing because it needs scipy/scipy#17691. What should we do - specify the derivative so that `method='newton'` can be used in old SciPy, or wait until SciPy 1.11 comes out before merging this PR?
I'd prefer to just add the derivative temporarily, then we can remove it when SciPy 1.11.0 is the minimum supported version of the lecture notes.

jarrodmillman:
> What should we do - specify the derivative so that `method='newton'` can be used in old SciPy, or wait until SciPy 1.11 comes out before merging this PR?

It looks like you specified the derivative in f0dd780. Would you remove that after  SciPy 1.11 is released?

jarrodmillman (approved):

mdhaber:
Yes

PR #590, opened 2023-05-26
ENH: scipy.optimize: tactfully communicate that algorithms aren't magic

mdhaber:


stefanv:
Best PR title!

stefanv:
Should we tell them about one of the "meta" optimizers that reinitialize guesses several times?

stefanv:
I'm thinking of https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.basinhopping.html and friends.

mdhaber:
I actually replaced the mention of `basinhopping` in favor of `differential_evolution` because the former has the lowest solve rate among SciPy's global optimizers (ignoring `brute`. See e.g. http://infinity77.net/go_2021/thebenchmarks.html, 10k function evaluation limit). If we only have space for one, I thought we should prioritize reliability and features over speed, and `differential_evolution` isn't a bad choice because it's pretty reliable and handles constraints.

stefanv:
What I meant was that, instead of only suggesting that users try new initial values that they choose themselves, perhaps point out that there are algorithms that can execute that process on their behalf.

jarrodmillman:
> What I meant was that, instead of only suggesting that users try new initial values that they choose themselves, perhaps point out that there are algorithms that can execute that process on their behalf.

@stefanv I am going to merge this, but it would be good to keep your comment above in an issue or something. Maybe we should make a todo list on hackmd, a wiki, a PR, or somewhere. That way we can collect plans, comments, ideas, etc. while we rapidly make minor changes.

PR #591, opened 2023-05-26
MAINT: fix equation rendering

tupui:


tupui:
Still some issues, looking at it. Locally it's not rendering for me.

jarrodmillman (approved):

PR #592, opened 2023-05-26
MAINT: update statistics section

mdhaber:
This adjusts the SciPy portion of the chapter, but I wouldn't be a good person to work on the rest.
Also, we might want to pick a different example or adjust the language surrounding the male/female distinction.

### DIFF
--- a/packages/statistics/index.rst
+++ b/packages/statistics/index.rst
@@ -314,16 +314,15 @@ use the :mod:`scipy.stats` sub-module of `SciPy
 Student's t-test: the simplest statistical test
 ------------------------------------------------
 
-1-sample t-test: testing the value of a population mean
+One-sample tests: testing the value of a population mean
 ........................................................
 
 .. image:: two_sided.png
    :scale: 50
    :align: right
 
-:func:`scipy.stats.ttest_1samp` tests if the population mean of data is
-likely to be equal to a given value (technically if observations are
-drawn from a Gaussian distributions of given population mean). It returns
+:func:`scipy.stats.ttest_1samp` tests the null hypothesis that the mean
+of the population underlying the data is equal to a given value. It returns
 the `T statistic <https://en.wikipedia.org/wiki/Student%27s_t-test>`_,
 and the `p-value <https://en.wikipedia.org/wiki/P-value>`_ (see the
 function's help)::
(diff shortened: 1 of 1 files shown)


packages/statistics/index.rst
```diff
@@ -314,40 +314,62 @@ use the :mod:`scipy.stats` sub-module of `SciPy
...
+One-sample tests: testing the value of a population mean
 ........................................................
 
 .. image:: two_sided.png
    :scale: 50
    :align: right
 
-:func:`scipy.stats.ttest_1samp` tests if the population mean of data is
-likely to be equal to a given value (technically if observations are
-drawn from a Gaussian distributions of given population mean). It returns
```
mdhaber, line 326: This may lead to two misunderstandings. I've adjusted the text here and added the paragraph beginning with "Technically" to avoid these.

packages/statistics/index.rst
```diff
@@ -314,40 +314,62 @@ use the :mod:`scipy.stats` sub-module of `SciPy
...
-    With a p-value of 10^-28 we can claim that the population mean for
-    the IQ (VIQ measure) is not 0.
+    >>> sp.stats.wilcoxon(data['VIQ'])  # doctest: +ELLIPSIS
+    WilcoxonResult(statistic=0.0, pvalue=1.8189894...e-12)
 
-2-sample t-test: testing for difference across populations
+Two-sample t-test: testing for difference across populations
 ...........................................................
 
-We have seen above that the mean VIQ in the male and female populations
```
mdhaber, line 342: We saw that the sample means are different. We want to make an inference about the populations.

packages/statistics/index.rst
```diff
@@ -314,40 +314,62 @@ use the :mod:`scipy.stats` sub-module of `SciPy
...
-.. tip::
+The p-value of :math:`10^-28` indicates that such an extreme value of the statistic
+is unlikely to be observed under the null hypothesis. This may be taken as
+evidence that the null hypothesis is false and that the population mean IQ
+(VIQ measure) is not 0.
+
+Technically, the p-value of the t-test is derived under the assumption that
+the means of samples drawn from the population are normally distributed.
+This condition is exactly satisfied when the population itself is normally
+distributed; however, due to the central limit theorem, the condition is
```
mdhaber, line 341: I've seen this written before, but it seems to me that it is safest to say that the sample t-statistic is distributed according to the t-distribution. I'm not sure whether this is guaranteed under the assumption stated above.

jarrodmillman (approved):

jarrodmillman:
I am merging this as it seems like a clear improvement. We can address additional issues in future PRs. Thanks!!

Issue #587, opened 2023-05-25
Hosting


tupui:
Since #574 got in, we can think about updating the hosting. I propose to do:

https://lectures.scientific-python.org

There is also a question about migrating the repository to Scientific Python's org.

GaelVaroquaux:
> https://lectures.scientific-python.org

If we do this, we should do a permanent redirect (301 code https://en.wikipedia.org/wiki/HTTP_301) from http://scipy-lectures.org/, so that the search engine optimization and bookmarks carry over.

G

tupui:
+1 to this, thank you Gael! I suppose we could also add a banner for folks using the old URL.

stefanv:
xref: #656

tupui:
Closing, all is done 🎉 Thank you folks =)

PR #593, opened 2023-05-27
Add client-side redirect to lectures.scientific-python.org

stefanv:
Since we're hosting on GitHub Pages, we don't have the option of using a 301 Moved Permanently.

But this is a route recommended by Google:

https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls https://developers.google.com/search/docs/crawling-indexing/301-redirects#metarefresh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,9 @@ html:
 	@echo
 	@echo "Build finished. The HTML pages are in build/html."
 
+html-scipy: export DOMAIN=scipy-lectures.org
+html-scipy: html
+
 cleandoctrees:
 	rm -rf build/doctrees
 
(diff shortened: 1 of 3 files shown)


stefanv:
This is in response to @GaelVaroquaux's comment in https://github.com/scipy-lectures/scipy-lecture-notes/issues/587#issuecomment-1563923769

@jarrodmillman The new site is live, but we do not have to move, and I can disable it for the time being if more time is needed. But, if we want to make the move, we need to deploy a new version of the docs with the redirection headers, as implemented in this PR.

We should either do that soon, or take the other site down again, so that Google doesn't consider these as duplicate sites.

jarrodmillman:
I would prefer to wait a bit. I was hoping to recruit a new team of authors and editors first. I also wanted to have a concrete plan for how we are going to teach with these materials. There are a few other things I wanted to do before the move.  Let's talk before deciding to make this live.

jarrodmillman (approved):

PR #597, opened 2023-06-07
Use jquery

jarrodmillman:


stefanv:
This is required by folding menus and copybutton.

Issue #568, opened 2023-05-20
Revisit scikit-image io section


jarrodmillman:
In #567 , I added the `# doctest: +SKIP` to the end of the following doctest:
```
Input/output, data types and colorspaces
=========================================

I/O: :mod:`skimage.io` ::

    >>> from skimage import io


Reading from files: :func:`skimage.io.imread` ::

    >>> import os
    >>> filename = os.path.join(skimage.data_dir, 'camera.png')
    >>> camera = io.imread(filename) # doctest: +SKIP
```
because without it (after installing pooch) we were getting:
```
=================================== FAILURES ===================================
_____________________________ [doctest] index.rst ______________________________
152 I/O: :mod:`skimage.io` ::
153 
154     >>> from skimage import io
155 
156 
157 Reading from files: :func:`skimage.io.imread` ::
158 
159     >>> import os
160     >>> filename = os.path.join(skimage.data_dir, 'camera.png')
161     >>> camera = io.imread(filename)
UNEXPECTED EXCEPTION: FileNotFoundError("No such file: '/home/runner/.cache/scikit-image/0.21.0rc1/data/camera.png'")
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/doctest.py", line 1350, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest index.rst[22]>", line 1, in <module>
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages/skimage/io/_io.py", line 53, in imread
    img = call_plugin('imread', fname, plugin=plugin, **plugin_args)
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages/skimage/io/manage_plugins.py", line 205, in call_plugin
    return func(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages/skimage/io/_plugins/imageio_plugin.py", line 11, in imread
    out = np.asarray(imageio_imread(*args, **kwargs))
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages/imageio/v3.py", line 53, in imread
    with imopen(uri, "r", **plugin_kwargs) as img_file:
  File "/opt/hostedtoolcache/Python/3.10.11/x64/lib/python3.10/site-packages/imageio/core/imopen.py", line 113, in imopen
[...]

PR #610, opened 2023-06-08
Review skimage module up to `io`

stefanv:
- Clean up library description and related libraries
- Avoid using `skimage.io.data_dir` by saving an example image to disk
- Use `import skimage as ski` convention


jarrodmillman:
Close #568.

jarrodmillman (approved):

PR #623, opened 2023-06-28
Use normal approximation

### LINKED ISSUE #619: Exact p-value calculation does not work if there are zeros.
We are getting the following warning on `main`:
```
packages/statistics/index.rst::index.rst
  /opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/scipy/stats/_morestats.py:4088: UserWarning: Exact p-value calculation does not work if there are zeros. Switching to normal approximation.
    warnings.warn("Exact p-value calculation does not work if there are "
```

@mdhaber Thoughts?

jarrodmillman:
Close #619.

### DIFF
--- a/packages/statistics/index.rst
+++ b/packages/statistics/index.rst
@@ -408,7 +408,7 @@ observations::
 Accordingly, we can perform a nonparametric version of the test with
 ``wilcoxon``.
 
-    >>> sp.stats.wilcoxon(data['FSIQ'], data['PIQ'])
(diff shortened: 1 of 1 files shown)


mdhaber (approved):
This is fine to prevent the warning. Explicit use of `permutation_test` to get an exact (or at least accurate) p-value even in case of ties is a more advanced topic for another time. And TBH we will probably begin to allow https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.PermutationMethod.html#scipy.stats.PermutationMethod soon enough. We can update this after that.

PR #630, opened 2023-06-28
Avoid conversion of array to scalar warning

jarrodmillman:
Avoid
```
intro/scipy/index.rst: 4 warnings
packages/statistics/index.rst: 32 warnings
  /opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/scipy/stats/_stats_py.py:1069: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    else dtype(mean))
```


stefanv:
See https://github.com/scipy/scipy/issues/18781

jarrodmillman:
See #631 , which also removes the warnings.

PR #641, opened 2023-06-29
Autoformat python scripts

jarrodmillman:


stefanv (approved):
We should eventually transition all pycon snippets to code-blocks.

PR #655, opened 2023-06-30
Prevent "X is already assigned section numbers" warnings

stefanv:
`.. only:: latex` and `.. only:: html` are both parsed when compiling, thus it cannot be used to "branch" table of contents for the two output formats.

Instead, we switch to `.. only:: html` for selecting the headers.


stefanv:
I have tried to check this, but I don't have super high confidence that it is correct.

jarrodmillman (approved):

PR #720, opened 2024-02-23
Update year

jarrodmillman:
$ sudo dnf install adobe-source-sans-pro-fonts lato-fonts


stefanv:
What is the font issue you are seeing? The diff on GH looks OK.

PR #771, opened 2024-06-27
Update scipy

jarrodmillman:


jarrodmillman:
@dschult @stefanv Just double-checking the repr change was intentional (see 8c8664f94ca1143b2f8907c2aa63872ff7441c6a).

dschult:
Yes — the repr change was intentional (and str change). 
It was driven by trying to fit 1D into the old format, but it
turned out to be a better change than we expected in other ways too.
;)

PR #785, opened 2024-07-28
Fix icons width

### LINKED ISSUE #784: Icons too large
Recently the front page changed from
2024-07-28T04:44:42,432029352-07:00
to this
2024-07-28T04:44:57,215350330-07:00

The only difference is that we were using Sphinx 7.3.7 before and now we are using 7.4.4.

The `width: 15em` is coming from
```
.. |pdf-icon| image:: images/icon-pdf.svg
   :width: 15em
   :class: vcenter
   :alt: PDF icon
```
in `index.rst`, which was added in #653.

jarrodmillman:
Fix #784

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -70,18 +70,15 @@ Scientific Python Lectures
 .. Icons from https://fonts.google.com/icons
 
 .. |pdf-icon| image:: images/icon-pdf.svg
-   :width: 15em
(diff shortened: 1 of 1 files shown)


jarrodmillman:
@stefanv Do you have any suggestions on how we should fix this?

PR #795, opened 2024-09-18
Fix value error: 0.7 -> 0.5

nobucshirai:
The function value mentioned in the description before the program in the section "13.2.1 Getting Started: 1D Optimization" differs from the value used in the program. To minimize changes, I updated the value of 'a' in the function f(x) = exp[(x - a)^2] from 0.7 to 0.5 in the description.


stefanv (approved):
Thanks @nobucshirai!

PR #810, opened 2024-12-09
Update NumPy (2.2)

jarrodmillman:
See https://github.com/numpy/numpy/issues/27957


jarrodmillman:
Still getting
```
advanced/optimizing/ica.py:90: error: Value of type variable "SupportsRichComparisonT" of "max" cannot be "floating[Any]"  [type-var]
Found 1 error in 1 file (checked 202 source files)
```

PR #817, opened 2025-01-10
Update scipy

jarrodmillman:
@mdhaber We are getting this error with 1.15.0 and 1.15.1
```
367     >>> sp.stats.wilcoxon(data['VIQ'])
Expected:
    WilcoxonResult(statistic=np.float64(0.0), pvalue=np.float64(1.8189894...e-12))
Got:
    WilcoxonResult(statistic=np.float64(0.0), pvalue=np.float64(3.488172636231201e-08))

/home/runner/work/scientific-python-lectures/scientific-python-lectures/packages/statistics/index.rst:367: DocTestFailure
```

Thoughts?

### DIFF
--- a/advanced/mathematical_optimization/index.rst
+++ b/advanced/mathematical_optimization/index.rst
@@ -887,11 +887,10 @@ Lets try to minimize the norm of the following vectorial function::
     >>> x0 = np.zeros(10)
     >>> sp.optimize.leastsq(f, x0)
     (array([0.        , 0.11111111, 0.22222222, 0.33333333, 0.44444444,
-           0.55555556, 0.66666667, 0.77777778, 0.88888889, 1.        ]), 2)
+           0.55555556, 0.66666667, 0.77777778, 0.88888889, 1.        ]), ...)
 
-This took 67 function evaluations (check it with 'full_output=1'). What
-if we compute the norm ourselves and use a good generic optimizer
-(BFGS)::
+This took 67 function evaluations (check it with 'full_output=True'). What
+if we compute the norm ourselves and use a good generic optimizer (BFGS)::
 
     >>> def g(x):
     ...     return np.sum(f(x)**2)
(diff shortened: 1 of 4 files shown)


mdhaber:
I discussed this with Stefan the other day.

>  The different result with SciPy 1.15.0 comes from the fact that we switched the method that is used automatically in case of ties. It used to be the "exact" method, which is by no means "exact" in case of ties.

So I think it would be fine to change the reference value.

jarrodmillman:
We are sometimes seeing this error now
```
884     >>> def f(x):
885     ...     return np.arctan(x) - np.arctan(np.linspace(0, 1, len(x)))
886 
887     >>> x0 = np.zeros(10)
888     >>> sp.optimize.leastsq(f, x0)
Expected:
    (array([0.        , 0.11111111, 0.22222222, 0.33333333, 0.44444444,
           0.55555556, 0.66666667, 0.77777778, 0.88888889, 1.        ]), 2)
Got:
    (array([0.        , 0.11111111, 0.22222222, 0.33333333, 0.44444444,
           0.55555556, 0.66666667, 0.77777778, 0.88888889, 1.        ]), 4)

/Users/runner/work/scientific-python-lectures/scientific-python-lectures/advanced/mathematical_optimization/index.rst:888: DocTestFailure
```

stefanv:
According to the docstring:

```
ier : int
    An integer flag. If it is equal to 1, 2, 3 or 4, the solution was
    found. Otherwise, the solution was not found. In either case, the
    optional output variable 'mesg' gives more information.
```

So, we could set `full_output` to True to see what the message is.

jarrodmillman:
Not sure why, but now the stopping criteria for leastsq is either 2 (relative error between iterations) or 4 (cosine of the angle). See https://github.com/scipy/scipy/blob/df134eab5a500c2146ed4552c8674a78d8154ee9/scipy/optimize/_minpack_py.py#L451 for more details.

stefanv (approved):

PR #834, opened 2025-05-06
Fix typo from ploints to points in plot_separator.py

tkoyama010:
Fix typo from ploints to points in plot_separator.py.


stefanv (approved):

stefanv:
Thank you!

PR #840, opened 2025-07-29
Update index.rst 1.5.7.2 equation

masonada2:
In 1.5.7.2. the equations dy/dy=... and dy_dot/dt=... are displayed connected side by side. I know nothing about git and text formatting so I've just put enter between equations.  

1 5 SciPy error


rossbar (approved):
Thanks @masonada2 , it does indeed look like a formatting issue and the added blank line fixes it!

jarrodmillman (approved):
Thanks!

